PR #1, opened 2017-09-17
DOC: how to run MPI programs

stsievert:
I'm not sure if you accept PRs via GitHub, but here it is. I'm just starting with mpi4py and spent a while looking for this.


dalcinl:
We accept PR from Bitbucket, but I could just grab the raw commit and apply it to master.

However, I don't like your use of `mpirun`. The standardized executor command is `mpiexec` using the `-n` flag. Could you please update your commit? Also the wording *running an MPI program* may be confusing, maybe is better to say *running a Python script*?

stsievert:
PR submitted on BitBucket: https://bitbucket.org/mpi4py/mpi4py/pull-requests/11/doc-running-python-scripts-with-mpi/diff

PR #3, opened 2019-12-03
Add note about pip using cache

ejjordan:
I did not see a way to open an issue and found this was a hiccup for me when installing multiple versions of mpi4py on my system. If the language needs some work I would be happy to rewrite but I think this note would be helpful for users.


dalcinl:
The git repository at GitHub is a just mirror, issues are disabled, but I cannot disable PRs. 
The official repo is at https://bitbucket.org/mpi4py/mpi4py.
Please submit your PR to the official repo.

Your wording is a bit inaccurate, I would replace `to renew the paths to the library objects::` by 
`to not use cached wheel files from previous builds::`.

dalcinl:
Please note pypa/pip#6391. Once the `pip cache ...` command is available, I guess that the best way to workaround the issue is to just `pip cache remove mpi4py`.

dalcinl:
Closed via 384bbf8.

PR #12, opened 2021-01-04
Python: Update code to Python 3, use lib3to2 to support Python 2

dalcinl:


dalcinl:
@leofang Looks like it is relatively safe to upgrade Python source code to Python 3 syntax. The 3to2 fixers produce rather minimal changes on the Python 3 sources. I reviewed carefully the output Python 2 code, and everything looks good.

leofang:
@dalcinl This is great! Glad it worked out. Are we one step closer toward a new release? 😇 Do we need to update the syntax in Cython files as well?

dalcinl:
> Do we need to update the syntax in Cython files as well?

No, no need to update Cython. I have a branch (in the Bitbucket repo) where I'm working on adding type hints with annotations.

PR #20, opened 2021-01-28
Type annotations and support for type checkers

dalcinl:


leofang:
Wow I forgot to ask: @dalcinl Did you add all these type annotations by hand, or there're some automation tools?

dalcinl:
All by hand! It is Cython code, I don't know any tool that would help with that.

PR #33, opened 2021-03-16
mpi4py.util.pkl5: Add support for `bcast()`

dalcinl:


dalcinl:
@hzhou Looks like Homebrew recently upgraded its formula to MPI 3.4.1 using `ch4:ofi`. I'm seeing tons of regressions in mpi4py's testsuite. I'm seeing the same problems my Linux workstation with my own `ch4:ofi` build, either with 3.4.1 release tarball or main branch (maybe a week old). Looks like some of these issues are related to changes in behavior of the progress engine. Have you guys got any similar report from other users?

Let's look at a reproducer using Python. This reproducer involves matched probes and receives, and a nonblocking send-to-self in synchronous mode. As it is, the code fails the second assertion. Uncommenting  the `#comm.Probe()` line, the snippet runs to completion. In understand that the MPI standard may not REALLY require this to work, but I think you would grant me that if you send to self, it is reasonable to expect that a subsequent probe for a message should match the previous send. In any case, that's  the behavior of the ch3 device, and this new one in ch4 may be an unintended regression.

```
from mpi4py import MPI
comm = MPI.COMM_SELF

msg = comm.Improbe(0, MPI.ANY_TAG)
assert msg is None

req = comm.Issend([b"@", 1, MPI.BYTE], 0, 1)
#comm.Probe()

msg = comm.Improbe(0, MPI.ANY_TAG)
assert msg is not None

buf = bytearray(2)
msg.Recv(buf)
assert buf[0:1] == b'@'

flag = req.Test()
assert flag
```

I'm still investigating other failures.

dalcinl:
@leofang This is a small one. The mayor pain was getting the testsuite updated for the new MPICH ch4 device. My ultimate goal is to add support for the new pickle 5 communication mode in `mpi4py.futures`.

conf/coverage.sh
```diff
@@ -54,6 +54,8 @@ $MPIEXEC -n 1 $PYTHON -m coverage run -m mpi4py -c "import sys; sys.exit('error'
 $MPIEXEC -n 1 $PYTHON -m coverage run -m mpi4py -c "from mpi4py import MPI; 1/0;"  > /dev/null 2>&1 || true
 
 $MPIEXEC -n 1 $PYTHON -m coverage run test/test_util_pkl5.py  -q 2> /dev/null
+$MPIEXEC -n 2 $PYTHON -m coverage run test/test_util_pkl5.py  -q 2> /dev/null
+$MPIEXEC -n 3 $PYTHON -m coverage run test/test_util_pkl5.py  -q 2> /dev/null
```
leofang, line 58: Just curious: `-n 2` is not enough?

demo/futures/test_futures.py
```diff
@@ -1305,6 +1305,9 @@ def transform(a):
     if version == (4,0,2) and sys.platform=='darwin':
         SKIP_POOL_TEST = True
 if name == 'MPICH':
+    if sys.platform == 'darwin':
+        if version >= (3, 4) and version < (4, 0):
+            SKIP_POOL_TEST = True
```
leofang, line 1310: Is this due to https://github.com/pmodels/mpich/issues/5149?

test/test_util_pkl5.py
```diff
@@ -593,6 +593,67 @@ def testMessage(self):
         self.assertEqual(msg, self.MessageType(MPI.MESSAGE_NO_PROC))
         self.assertNotEqual(msg, MPI.MESSAGE_NULL)
 
+    def testBcastIntra(self, msglist=None, check=None):
+        comm = self.COMM
+        size = comm.Get_size()
+        for smess in (msglist or messages):
```
leofang, line 599: I don't think `messages` gets tested, only `msglist`? Is this intended?

leofang (approved):
Thanks, @dalcinl! I skimmed over most of the changes except for `testBcastInter`, as I'm in a remote conference now and can only have fun on spare time. 

LGTM, just left a few questions, but do not let me block you from merging this 🙂

test/test_util_pkl5.py
dalcinl: No, it is tested. At some point the `unitest` infrastructure will call `testcase.testBcastIntra()` without arguments, and then the `messages` case will be tested.

demo/futures/test_futures.py
dalcinl: Yes. Current MPICH 3.4.1 from Homebrew on macOS Catalina does not work with the `mpi4py.futures` tests that involve `MPI_Comm_spawn()`. I have no idea what's going on, I cannot reproduce the problem in Linux. So I just disabled the test hoping for future resolution in MPICH 4.0.

conf/coverage.sh
dalcinl: No, if you want to test `bcast()` on intercommunicators, and exercise both the `MPI.ROOT` and `MPI.PROC_NULL` branches. So you need a universe of 2+1 processes to be rearranged as a 2<->1 intercommunicator. So 3 processes is the minimum required to achieve full test coverage.

leofang: Thanks for explanation, Lisandro. Yeah it was a silly question if I had read the `testBcastInter` test 😅

test/test_util_pkl5.py
leofang: I misread something, and also I forgot it's part of `unittest`, so this function would be called with the default args indeed (since it has a prefix `test`).

PR #35, opened 2021-03-20
Update mpi.cfg for Fujitsu MPI

dalcinl:
Add support for Fujitsu MPI.

### DIFF
--- a/mpi.cfg
+++ b/mpi.cfg
@@ -192,3 +192,13 @@ include_dirs = conf/nompi
 # --------------
 [mpiuni]
 include_dirs = conf/mpiuni:$PETSC_DIR/include:$PETSC_DIR/$PETSC_ARCH/include
+
+
+# Fujitsu MPI example
+# -------------------
+[fujitsu-mpi]
+mpicc                = mpifcc
+mpicxx               = mpiFCC
+define_macros        = OPENMPI_DLOPEN_LIBMPI=1
+extra_compile_args   = -Nclang
+extra_link_args      = -Knolargepage
--- a/src/lib-mpi/compat/openmpi.h
+++ b/src/lib-mpi/compat/openmpi.h
@@ -65,7 +65,9 @@ static void PyMPI_OPENMPI_dlopen_libmpi(void)
   mode |= RTLD_NOLOAD;
   #endif
   #if defined(OMPI_MAJOR_VERSION)
-  #if OMPI_MAJOR_VERSION == 3
+  #if   OMPI_MAJOR_VERSION >= 4
+  if (!handle) handle = dlopen("libmpi.40.dylib", mode);
+  #elif OMPI_MAJOR_VERSION == 3
   if (!handle) handle = dlopen("libmpi.40.dylib", mode);
   #elif OMPI_MAJOR_VERSION == 2
   if (!handle) handle = dlopen("libmpi.20.dylib", mode);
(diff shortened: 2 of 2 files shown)


dalcinl:
@ggouaillardet We are slowly moving to GitHub, although it is not official yet. Thanks for your original PR.

I cherry-picked your original commit to this PR. 

I hoped that the need for this hack was a thing of the past, but it looks like that the only way to get things fixed once and for all is to add a test to Open MPI testsuite explicitly excising `dlopen()` with `RTLD_LOCAL`. I have to add a fix for the dlopen() hacks to handle the v4.x series. You were lucky, the `OPENMPI_DLOPEN_LIBMPI=1` trick worked out of the box either because Open MPI did not change the library version from v3.x to v4.x, otherwise the `dlopen()` breaks (unless you have installed the unversioned `libmpi.so` symbolic link)

ggouaillardet:
Thanks @dalcinl 

FWIW with Fujitsu MPI (that is Open MPI 4 based)

```
libmpi.so -> libmpi.so.0.0.0
libmpi.so.0 -> libmpi.so.0.0.0
libmpi.so.0.0.0
```

dalcinl:
Oh, boy, really? OK, so all this is working by luck. Is the `mpi.h` header available somewhere? Do they define some macro to advertise themselves as Fujitsu MPI to applications? For sure they define the OMPI_XXX macros, but is there anything specific?

ggouaillardet:
`MPI_Get_library_version()` returns
```
FUJITSU MPI Library 4.0.0 (4.0.1fj4.0.0)
```

but I did not spot anything specific in the `mpi.h` macros.


That being said, the following macros are defined in `mpi-ext.h`

```
OMPI_HAVE_MPI_EXT_FJCOLLECTION
OMPI_HAVE_MPI_EXT_FJMSWK
OMPI_HAVE_MPI_EXT_FJPREQUEST
OMPI_HAVE_MPI_EXT_FJPROGRESS
OMPI_HAVE_MPI_EXT_FJRUNTIME
OMPI_HAVE_MPI_EXT_FJTOPOLOGY
```

dalcinl:
> That being said, the following macros are defined in `mpi-ext.h`

However ``mpi-ext.h`` is not included by the standard `mpi.h`. Should I trust other Open MPI derivatives to always provide this extra header? I'm a bit hesitant, I still remember the Spectrum MPI folks messing with the OMPI version numbers to expose their own version.

What about asking the Fujitsu folks about this `dlopen()` issue and beg for a fix? Do you know whether they receptive to such requests? After getting a fix upstream for future releases, then I'm MUCH more willing to use my time implementing a proper,  robust workaround for previous releases.

ggouaillardet:
@dalcinl I will have this reported to Fujitsu, and hope they will fix it.

FWIW, the issue was more complex than anticipated...
It occurs because:
 - An external PMIx is used **and**
 - The external PMIx version is < `3.1.3` **and**
 - The external PMIx was **not** built with `--disable-dlopen` (or in embedded mode)

The fix is at openpmix/openpmix@facf38469803cc096322f8cb9f89feb4647f564a

Vendors are free to build Open MPI derivatives the way they want. That being said, I think removing `mpi-ext.h` would be an extra work with no value associated to it.

dalcinl:
OK, so the starts aligned to conspire with us. I believe the best course of action is to leave things as they are. As Fujitsu MPI installs the symlink (`libmpi.so -> libmpi.so.0.0.0`), mpi4py will `dlopen()` via `libmpi.so` as a last resort and things will work.

PR #39, opened 2021-03-30
mpi4py.futures: Fix sys.path[0] on workers if using `-m` flag

### LINKED ISSUE #38: mpi4py.futures unable to resolve imports of other modules in same package as main script
I'm trying to use the MPIPoolExecutor on a Beowulf HPC cluster running Maui and Torque. I have a Python project consisting of multiple files (so a package containing multiple modules). The project structure is basically like this:

```
my_project
  - __init__.py
  - my_main_script.py
  - my_other_module.py
```

Using the MPIPoolExecutor in the main script works fine until I want to import another module inside the same package.

This simple code inside `my_main_script.py` works without problems:
```
from mpi4py.futures import MPIPoolExecutor


def f(x):
    return x ** 2


if __name__ == '__main__':
    with MPIPoolExecutor() as executor:
        print(list(executor.map(f, range(10))))
```

To run the code, I used this command: `mpiexec -n 4 -rf my_rank_file python -m mpi4py.futures my_main_script.py`

However, let's say that the function `f` is imported from another file in the same folder, `my_other_module.py`. Then `my_main_script.py` looks like this:
```
from mpi4py.futures import MPIPoolExecutor
from my_other_module import f


if __name__ == '__main__':
    with MPIPoolExecutor() as executor:
        print(list(executor.map(f, range(10))))
```

When I run this with the same command, I get the following error: `ModuleNotFoundError: No module named 'my_other_module'`

The error only seems to occur on worker nodes, not on the master node.

I'm using Open MPI 1.8.8, mpi4py 3.0.3, and Python 3.8.7.

dalcinl:
Fixes #38.

### DIFF
--- a/src/mpi4py/futures/_lib.py
+++ b/src/mpi4py/futures/_lib.py
@@ -728,6 +728,8 @@ def import_main(mod_name, mod_path, init_globals, run_name):
         if mod_name:  # pragma: no cover
             runpy.run_module(mod_name, run_name=run_name, alter_sys=True)
         elif mod_path:  # pragma: no branch
+            if not getattr(sys.flags, 'isolated', 0):  # pragma: no branch
+                sys.path[0] = os.path.realpath(os.path.dirname(mod_path))
             runpy.run_path(mod_path, run_name=run_name)
         sys.modules['__main__'] = sys.modules[run_name] = module
     except:  # pragma: no cover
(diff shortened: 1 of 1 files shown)


dalcinl:
@adriaanpardoel Just FYI, this is the proper fix that covers both Python 3 and legacy Python 2. Note that my previous patch in #38 is ok for Python 3. If you have a Python 2 interpreter around in your cluster and you could perform a quick test with some trivial code printing `sys.path[0]`, I would really appreciate it.

adriaanpardoel:
I ran a test with Python 2 on the cluster now. With the current pypi release, it gives me `sys.path[0]` is `''` on worker nodes. After installing the version on this branch, it's now `'/absolute/path/to/script/directory'`, so seems to work!

dalcinl:
> so seems to work!

Many thanks for the confirmation! Merging...

PR #40, opened 2021-03-31
test: Alignment of send/receive buffers and data types in test_rma

saosudo:
In this pull request, I fixed a violation of the MPI standard in the way MPI_Compare_and_swap and MPI_Fetch_and_op are called in test_rma.py.
 
I fixed it like a pull request and it now passes, but I feel like it's weird to fix the disp, so do you know the right way to do it?
 
I think it is essential that the alignment of the send and receive buffers match the alignment of the data types.
 
In MPI-3.1, 3.3.1 Type Matching Rule, it says
```
Type matching has to be observed at each of these three phases: 
The type of each variable in the sender buffer has to match the type specified for that entry by the send operation; 
the type specified by the send operation has to match the type specified by the receive operation; 
and the type of each variable in the receive buffer has to match the type specified for that entry by the receive operation. 
A program that fails to observe these three rules is erroneous.
```
 
This means that if you use MPI_INT, the send/receive buffer must be an int variable, and a program that uses a send/receive buffer that is not 4-byte aligned would be wrong.
 
For example, MPI_BYTE is 1 byte, so there is no problem with this test program, but in other cases, we need to multiply the size of the data type.
 
Please point out if I have misread the MPI standard.
 
I'm having a trouble with Fujitsu MPI because this is causing errors in this test, so please fix this.

### DIFF
--- a/test/test_rma.py
+++ b/test/test_rma.py
@@ -161,6 +161,7 @@ class BaseTestRMA(object):
                         self.assertEqual(gbuf[-1], -1)
 
     def testFetchAndOp(self):
+        typemap = MPI._typedict
         group = self.WIN.Get_group()
         size = group.Get_size()
         rank = group.Get_rank()
@@ -187,6 +188,7 @@ class BaseTestRMA(object):
             if typecode in 'FDG': continue
             obuf = array(+1, typecode)
             rbuf = array(-1, typecode, 2)
+            datatype = typemap[typecode]
             for op in (MPI.SUM, MPI.PROD,
                        MPI.MAX, MPI.MIN,
                        MPI.REPLACE, MPI.NO_OP):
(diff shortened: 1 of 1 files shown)


dalcinl:
@saosudo I believe your interpretation of the standard is totally right. Thanks.

dalcinl (approved):

dalcinl:
> Please point out if I have misread the MPI standard.

But why would then the MPI standard allow you to create struct datatypes with misaligned entries? In any case, I agree with your common sense constratins to support efficient compare-and-swap and fetch-and-op, so I'm happy to merge this PR.

dalcinl:
I have a few questions/suggestions about Fujistsu MPI. For example, I would like to fix `MPI.get_vendor()` to return `"Fujitsu MPI"`. What's the best channel to ask a few basic questions? I'm not a user looking for support, but a developer trying to improve the interoperability of your software and my software, for the benefit of your users.

saosudo:
I can't find any contact information for Fujitsu MPI on GitHub, so I'm in the process of inquiring internally for an appropriate contact, so please wait a day or two.

nariba-fj:
@dalcinl  
Hi, I'm a colleague of @saosudo and I'm in charge of Fujitsu MPI development.
Since there is no repository for Fujitsu MPI and there is no appropriate place to contact you on GitHub, if that is not a problem, I will contact you individually next week for your email address.

dalcinl:
@nariba-fj Yes, please contact me via email. Alternatively, you can open issue in mpi4py/mpi4py.

nariba-fj:
@dalcinl Okay. I'll create an issue in this repository.

PR #92, opened 2021-09-11
mpi4py.util.dtlib: Fix alignment

### LINKED ISSUE #91: TestUtilDTLib fails on s390x: incorrect long double alignment
We discussed at https://github.com/mpi4py/mpi4py/issues/80#issuecomment-902349139 a test failure on s390x architecture:
```
FAIL: testStruct4 (test_util_dtlib.TestUtilDTLib) (typecode='g')
  File "/<<PKGBUILDDIR>>/test/test_util_dtlib.py", line 165, in testStruct4
    self.assertEqual(mt.extent, n*ex1)
AssertionError: 40 != 48

======================================================================
FAIL: testStruct4 (test_util_dtlib.TestUtilDTLib) (typecode='G')
  File "/<<PKGBUILDDIR>>/test/test_util_dtlib.py", line 165, in testStruct4
    self.assertEqual(mt.extent, n*ex1)
AssertionError: 72 != 80
```

Tests on s390x showed this:
```
#include <stddef.h>
#include <stdio.h>

typedef struct {char a; long double b; char c;} s;

int main() { 
  printf("long double has alignment %d\n", offsetof(s,b));
  printf("long double has size %d\n", sizeof(long double));
  printf("stuct s has size %d\n", sizeof(s));
}
```
gives
```
$ ./mtest2
long double has alignment 8
long double has size 16
stuct s has size 32
```

In python,
```
# import numpy as np                                                                                                                    
# print(np.dtype('c,g,c', align=True))
{'names':['f0','f1','f2'], 'formats':['S1','>f16','S1'], 'offsets':[0,8,24], 'itemsize':32, 'aligned':True}
```

and also in python,
```
from ctypes import *

class S(Structure):
    _fields_ = [
        ("a", c_char),
        ("b", c_longdouble),
        ("c", c_char),
    ]

print(sizeof(S))
```
returns 32.

The alignment of long double on s390x is not the same as that expected by mpi4py as found on other architectures.

dalcinl:
Do not assume alignment, use NumPy or ctypes to determine struct slot offsets.

Closes #91.

### DIFF
--- a/src/mpi4py/util/dtlib.py
+++ b/src/mpi4py/util/dtlib.py
@@ -24,22 +24,67 @@ def _get_typecode(datatype):
     return MPI._typecode(datatype)
 
 
+def _get_alignment_ctypes(typecode):
+    # pylint: disable=protected-access
+    # pylint: disable=import-outside-toplevel
+    import ctypes as ct
+    if typecode in ('p', 'n', 'P', 'N'):
+        kind = 'i' if typecode in ('p', 'n') else 'u'
+        size = ct.sizeof(ct.c_void_p)
+        typecode = '{}{:d}'.format(kind, size)
+    if typecode in ('F', 'D', 'G'):
+        typecode = typecode.lower()
+    if len(typecode) > 1:
+        mapping = {
+            'b': (ct.c_bool,),
+            'i': (ct.c_int8, ct.c_int16, ct.c_int32, ct.c_int64),
+            'u': (ct.c_uint8, ct.c_uint16, ct.c_uint32, ct.c_uint64),
+            'f': (ct.c_float, ct.c_double, ct.c_longdouble),
+        }
+        kind, size = typecode[0], int(typecode[1:])
+        if kind == 'c':
+            kind, size = 'f', size // 2
+        for c_type in mapping[kind]:
+            if ct.sizeof(c_type) == size:
+                typecode = c_type._type_
+    c_type_base = ct._SimpleCData
+    c_type = type('c_type', (c_type_base,), dict(_type_=typecode))
+    fields = [('base', ct.c_char), ('c_type', c_type)]
+    struct = type('S', (ct.Structure,), dict(_fields_=fields))
+    return struct.c_type.offset
+
+
+def _get_alignment(datatype):
+    typecode = _get_typecode(datatype)
+    if typecode is None:
+        combiner = datatype.combiner
+        combiner_f90 = (
+            MPI.COMBINER_F90_INTEGER,
+            MPI.COMBINER_F90_REAL,
+            MPI.COMBINER_F90_COMPLEX,
+        )
+        if combiner in combiner_f90:
+            typesize = datatype.Get_size()
+            typekind = 'ifc'[combiner_f90.index(combiner)]
+            typecode = '{0}{1:d}'.format(typekind, typesize)
+    if typecode is None:
+        # pylint: disable=import-outside-toplevel
+        from struct import calcsize
+        alignment = datatype.Get_size()
+        return min(max(1, alignment), calcsize('P'))
+    try:
+        return _np_dtype(typecode).alignment
+    except NameError:  # pragma: no cover
+        return _get_alignment_ctypes(typecode)
+
+
 def _is_aligned(datatype, offset=0):
     """Dermine whether an MPI datatype is aligned."""
     if datatype.is_predefined:
         if offset == 0:
             return True
-        typecode = _get_typecode(datatype)
-        combiner = datatype.combiner
-        is_complex = any((
-            typecode in ('F', 'D', 'G'),
-            typecode in ('c4', 'c8', 'c16', 'c32'),
-            combiner == MPI.COMBINER_F90_COMPLEX,
-        ))
-        typesize = datatype.Get_size() or 1
-        if is_complex:
-            typesize //= 2
-        return offset % typesize == 0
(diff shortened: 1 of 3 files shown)


drew-parsons:
No, must have missed something :(
https://buildd.debian.org/status/fetch.php?pkg=mpi4py&arch=s390x&ver=3.1.1-3&stamp=1631401483&raw=0
```
FAIL: testStruct4 (test_util_dtlib.TestUtilDTLib) (typecode='g')
  File "/<<PKGBUILDDIR>>/test/test_util_dtlib.py", line 168, in testStruct4
    self.assertTrue(dt.isalignedstruct)
AssertionError: False is not true

======================================================================
FAIL: testStruct4 (test_util_dtlib.TestUtilDTLib) (typecode='G')
  File "/<<PKGBUILDDIR>>/test/test_util_dtlib.py", line 168, in testStruct4
    self.assertTrue(dt.isalignedstruct)
AssertionError: False is not true
```

dalcinl:
> `dt` must also also be irregular,

No, no. It is still a problem of my implementation, I'm assuming alignment is equal to the type size
https://github.com/mpi4py/mpi4py/blob/maint/src/mpi4py/util/dtlib.py#L39-L42

The builtin `struct` module does not support `long double`. And `mpi4py.util.dtlib` is coded to still be semi-useful when NumPy is not available. I don't know how to compute `long double` alignment robustly in all scenarios. Perhaps using `ctypes`, but is is an option I would like to avoid, I'm not a big fan of it, and I need a bunch of code lines just to map typecode -> ctypes C type name.

drew-parsons:
s390x now passes tests successfully with the updated patch,
https://buildd.debian.org/status/fetch.php?pkg=mpi4py&arch=s390x&ver=3.1.1-4&stamp=1631618262&raw=0

dalcinl:
@drew-parsons If you do not have further concerns or comments, I'll merge this one. Please confirm.

drew-parsons:
Fine by me to merge.  The page is looking green.  https://buildd.debian.org/status/package.php?p=mpi4py&suite=experimental

dalcinl:
What about [riscv64](https://buildd.debian.org/status/fetch.php?pkg=mpi4py&arch=riscv64&ver=3.1.1-4&stamp=1631632239&raw=0)?

drew-parsons:
riscv64 is a niche architecture, its failure won't hold up the project.  The failure is curious, MPI itself seems to be hanging.  My guess is that it's running in a cross-platform VM, and openmpi is not so reliable inside the VM. ia64 will probably end up showing the same problem (been building for 5 hours now, it's probably been hung on the MPI test for 4 hours 50 minutes).

From my perspective we don't need to fix riscv64. Nice to support it if possible, but it's optional, not an obligation.

The tests do pass on single processor. Running under mpirun is where they fail. Of course mpi4py is an intrinsically mpi package, so it's a pity if they don't pass as MPI tests.

If the VM is the problem then it may well be that the riscv64 build is perfectly fine, it just needs real metal to run on. If you like, we could just disable tests on riscv64 (and ia64) and then the (untested) packages will be available for the niche users.

dalcinl:
> If you like, we could just disable tests on riscv64 (and ia64)

Could you instead do a scratch build and test using just two MPI processes, and passing `-fv` (fail-fast + verbose) to the runtests.py script? I would like to see which test is the one failing. Worst case, we can patch to disable the test as a know failure.

I'm merging this one now. Thanks for your help and patience.

drew-parsons:
We don't have a porterbox to test riscv64 alone, but I can make an upload to hit all the build daemons. Perhaps not a bad idea - the "standard" amd64 build just timed out in MPI the same way that riscv64 did.

drew-parsons:
Dropping back to only 2 processes for tests has let them untangle. All slots green now, https://buildd.debian.org/status/package.php?p=mpi4py&suite=experimental

Do you want to leave it at that, or want to dig a little deeper to see why they hang on 5 processes?

drew-parsons:
... I  spoke too soon, didn't wait for riscv64

dalcinl:
* The failure in riscv64 is unrelated to MPI. It looks like a network issue. This code uses client/server sockets within the builder, and depending on how network is configured, connections may fail. The failure is occurring before MPI has a chance of using these sockets.  I have not idea what would be the issue/configuration, and it is probably very hard to debug for you without access to a porterbox. I would suggest to patch the test method with the usual decorator `@unittest.skipIf(platform.machine() = '???', 'riscv64')`.
* Running with two MPI processes should be enough, although it is not ideal. Anyway, I believe that your responsibility is mpi4py, and  not hunt for and fix all the possible bugs from a backend MPI implementation. Perhaps you can update the scripts to lower the number of processes to 2 only for selected arches where you know things do  not work well or are too flaky to use 5? Or perhaps try 3 processes?

drew-parsons:
I'll do that.  I'll try using 5 processes for all arches except ia64 (amd64 sometimes breaks on 5 processes, but will succeed on a second attempt). I'll try 3 for ia64.  Perhaps riscv64 will be fine with 5 processes once the problematic test is skipped.

drew-parsons:
ia64 passes on 2 processes, but stalls on 3 processes in test_dynproc.TestDPM.testJoin.
risck64 passes fine on 5 processes after skipping testJoin.

Evidently testJoin is the central problem here.  I can keep the tests running on 5 processes, and skip testJoin for the problematic architectures. Seems to work fine that way.

PR #113, opened 2021-10-05
Allow adding an optional local version to the package version

maxhgerlach:
Hi,

we would like to build several variations of the same version of mpi4py and host them on an internal repository. It would be great to be able to add a local version to the package version so we can (1) make sure that we install the local binary package and not the public package and (2) distinguish between mpi4py binaries built for, e.g., multiple versions of MPI.

With this PR users can optionally define an environment variable `MPI4PY_LOCAL_VERSION` to append a local version identifier following https://www.python.org/dev/peps/pep-0440/#local-version-identifiers.


dalcinl:
Two things:

1.  I would like this PR to be merged into branch 'maint', so we have it working for the next patch release.

2. Would you mind doing it the following way? This would handle an empty value in the environ var, while our patch does not.

```diff
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,11 @@ def version():
     srcdir = os.path.join(topdir, 'src')
     with open(os.path.join(srcdir, 'mpi4py', '__init__.py')) as f:
         m = re.search(r"__version__\s*=\s*'(.*)'", f.read())
-        return m.groups()[0]
+    public_version = m.groups()[0]
+    local_version = os.environ.get('MPI4PY_LOCAL_VERSION')
+    if local_version:
+        return '{0}+{1}'.format(public_version, local_version)
+    return public_version
```

maxhgerlach:
Hi @dalcinl, yeah some thing went awfully wrong there; this was the first time that I tried to move a PR to a non-master branch... Sorry for the noise!

And thank you very much for checking this out and the very valid suggestion!

I will open a new PR.

PR #116, opened 2021-10-06
Allow adding an optional local version to the package version

maxhgerlach:
This is a rework of PR #113, this time based on branch `maint` and incorporating @dalcinl's suggestion to fully ignore `MPI4PY_LOCAL_VERSION` if it's set to an empty value.

-----

Hi,

we would like to build several variations of the same version of mpi4py and host them on an internal repository. It would be great to be able to add a local version to the package version so we can (1) make sure that we install the local binary package and not the public package and (2) distinguish between mpi4py binaries built for, e.g., multiple versions of MPI.

With this PR users can optionally define an environment variable `MPI4PY_LOCAL_VERSION` to append a local version identifier. They could follow the conventions lined out in https://www.python.org/dev/peps/pep-0440/#local-version-identifiers.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,12 @@ def version():
     srcdir = os.path.join(topdir, 'src')
     with open(os.path.join(srcdir, 'mpi4py', '__init__.py')) as f:
         m = re.search(r"__version__\s*=\s*'(.*)'", f.read())
-        return m.groups()[0]
+    public_version = m.groups()[0]
+    local_version = os.environ.get('MPI4PY_LOCAL_VERSION')
+    if local_version:
+        return '{0}+{1}'.format(public_version, local_version)
+    else:
+        return public_version
 
 def description():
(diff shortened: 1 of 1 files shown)


dalcinl:
@maxhgerlach We had an oversight in this PR: source (sdist) vs binary (wheel) builds.

1. Someone patching mpi4py sources may want to create an sdist and use the local version to state that fact.
2. And then someone else may take that patched sdist with local version, and want to build variant wheels (e.g. mpich vs openmpi).

Case (1) is slightly broken: creating a build from a sdist with local version simply discards the local version. This is because mpi4py gets its version number from `__init__.py`, and not the sdist metadata stored in `PKG-INFO`.

Case (2) may get slightly broken if we address case (1). Should the two (source and binary) local versions be concatenated (using a dot '.', source first and binary second) ?

Thoughts?

PR #130, opened 2021-11-01
mpi4py.util.pkl5: Use MPI4PY_PICKLE_THRESHOLD environment variable

dalcinl:


dalcinl:
@mrogowski Is this good enough for your needs?

mrogowski:
Looks good!

src/mpi4py/util/pkl5.py
```diff
@@ -62,10 +63,15 @@ def buf_cb(buf):
     return bufs, buf_cb
 
 
+def _get_threshold(default):
+    varname = 'MPI4PY_PICKLE_THRESHOLD'
+    return int(_os.environ.get(varname, default))
```
leofang, line 68: Just a nit: Can we have this value cached so we don't look it up every time?

dalcinl: Use the source, Luke! This function is executed only once, at import time, to define a class attribute.

leofang: Ah I misread it, LGTM then.

PR #131, opened 2021-11-03
Release/3.1.2

dalcinl:


leofang (approved):

dalcinl:
@drew-parsons @opoplawski FYI. I'll publish the new release by tomorrow, then you can start the usual packaging dance. Previous patches may no longer be needed.

drew-parsons:
Looking forward to it

leofang:
Will do. I'd like to add a section "Using conda" to `docs/source/usrman/install.rst`, above "Using pip"

dalcinl:
> above "Using pip"

Perhaps it should be done after, right before Testing? Otherwise, the "Requirements" section will look silly, you do not really require an MPI install when using conda and conda-forge binaries. 

PS: Will you mention the support for external MPICH and the additional steps to get Open MPI + GPU stuff working?

leofang:
@dalcinl Let's move the discussion to #132.

dalcinl:
@leofang I incorporated an edited version of your comments in `DESCRIPTION.rst`, which is what would be rendered on  [PyPI](https://pypi.org/project/mpi4py/) once I make the release. Eventually, all this stuff should be merged with the stuff under `docs/` to avoid repetition, but we need to find a clean way to do that.

@drew-parsons I've added a few notes about installing on Ubuntu/Debian in. I hope I got those right, otherwise let me know ASAP.

drew-parsons:
Hi @dalcinl , I can't see the apt notes. Can you point me in the right direction?

DESCRIPTION.rst
```diff
@@ -51,41 +51,105 @@ This package supports:
...
+
+  $ module load mpi/mpich-$(arch)
+  $ python -c "from mpi4py import MPI"
+
+* for the ``openmpi`` package do::
+
+  $ module load mpi/openmpi-$(arch)
+  $ python -c "from mpi4py import MPI"
+
+On **Ubuntu Linux** and **Debian Linux** systems, binary packages are
```
dalcinl, line 127: @drew-parsons Here are the `apt` notes. Please look at the next paragraph as well, about mpi4py being openmpi-only. Any suggestions?

drew-parsons: Of course, still in this PR.  I was looking in the main code!

The apt description is correct, and useful to identify that debian uses OpenMPI.
Could also mention docs are available,
"Local online documentation is available in the python-mpi4py-doc package."

With regards to using mpich on a Debian system. correct to note that libmpich-dev is needed. It is also possible (and "easy") to rebuild the debian package itself for mpich, just set the MPI variable in debian/rules to .mpich. That could help some library management problems, might introduce others. Someone choosing to use mpich on a Debian system would need to know what they're doing, in terms of impact on other packages.

Would it be interesting for you to enable joint installation of both openmpi and mpich builds at the same time?  The approach could be analogous to what I did with the serial and mpi builds of h5py, place specific implementations in subdirs (mpi4py/openmpi, mpi4py/mpich, also mpi4py/intelmpi etc) with toplevel __init__.py pulling in the preferred implementation. In principle that could be useful for other systems, not just Debian/Ubuntu. We can open a separate issue if we want to go down this path.

dalcinl: A mayor issue I have with the Ubuntu/Debian approach is that they use `alternatives` to switch from MPICH/Open MPI. Beside some issues with parallel installation that usually lead to users having broken environments (libs from one packge, mpiexec from another package)... Are users without sudo access able to switch from one implementation to the other? IMHO, a modules-based approach as Fedora uses is a more robust solution, although I understand it may be slightly harder to use/understand for newbies. Perhaps I'm getting it wrong, in such case, I apologize. 

In any case, giving users that chance to use an alternative implementation (to sort out bugs or third-party requirements) is always good. Also remember that MPICH and derivatives are ABI-compatible, so having an MPICH build that could work with an alternate MPICH-ABI installation makes a lot of sense.

drew-parsons:
Great, thanks Lisandro.
I'll open a separate Issue to discuss the question of joint MPI installation

leofang:
> @leofang I incorporated an edited version of your comments in `DESCRIPTION.rst`, which is what would be rendered on [PyPI](https://pypi.org/project/mpi4py/) once I make the release. Eventually, all this stuff should be merged with the stuff under `docs/` to avoid repetition, but we need to find a clean way to do that.

Thanks @dalcinl! It's kinda funny that one has to go to PyPI to read the installation instruction for conda 😄 but it's progress and we can improve it after 3.1.2 is out.

dalcinl:
@leofang You are definitely right. This was the quickest way I found to incorporate your changes. As I said before, all this information has to be consolidated under `docs/`.

PR #137, opened 2021-11-09
Add missing support for BOTTOM to generalized all-to-all collectives

dalcinl:


dalcinl:
@RolfRabenseifner I'm merging this one. The tests I've added cover the main two use cases for BOTTOM: a) datatypes with absolute addresses, and b) passing addresses via displacements (under MPI-3, this can only work with neighbor alltoallw, as it requires `MPI_Aint` displacements). All of the new tests pass, so I'm confident I got it right. In case I'm missing something, we can push additional fixes to the `maint` branch before the upcoming 3.1.3 release.

dalcinl:
Thanks @th-hlrs for the bug report.

PR #167, opened 2022-01-18
test: Remove dependency on distutils

dalcinl:


merwok:
I don’t think it’s correct to copy code from the Python standard library without giving attribution and copying the license.

It could be simpler and more robust to use a module that’s designed to be reused, like `packaging` which provides version parsing compliant with the PEPs.

PR #174, opened 2022-02-21
Add support for the MPI-4.0 standard

dalcinl:
* Use large count MPI-4 rutines.
* Add persistent collective communication.
* Add partitioned point-to-point communication.
* Add new communicator constructors.
* Add the `Session` class and its methods.


dalcinl:
@jeffhammond FYI. Now that MPICH-4.0 is out, it is time to follow suit.

src/mpi4py/MPI.pyi
```diff
@@ -538,11 +555,14 @@ class Info:
     def __contains__(self, value: str) -> bool: ...
     @classmethod
     def Create(cls) -> Info: ...
+    @classmethod
+    def Create_env(cls, args: Optional[Sequence[str]] = None) -> Info: ...
     def Free(self) -> None: ...
     def Dup(self) -> Info: ...
-    def Get(self, key: str, maxlen: int = -1) -> Optional[str]: ...
```
jeffhammond, line 543: I don't know if it matters at all, but you could use a tighter bound here:
MPI_MAX_INFO_KEY Maximum length of an info key
MPI_MAX_INFO_VAL Maximum length of an info value

test/test_errhandler.py
```diff
@@ -8,6 +8,43 @@ def testPredefined(self):
...
         self.assertTrue(MPI.ERRORS_ARE_FATAL)
         self.assertTrue(MPI.ERRORS_RETURN)
+        if MPI.VERSION >= 4:
+            self.assertTrue(MPI.ERRORS_ABORT)
+        else:
+            self.assertFalse(MPI.ERRORS_ABORT)
+
+    def testGetErrhandler(self):
+        errhdls = []
+        for i in range(100):
```
jeffhammond, line 18: Nothing in this loop depends on `i`.  What is it doing?

dalcinl: Perhaps I should have written `for _ in range(100):`, as the loop variable is indeed not used . The purpose of the test is to stress a bit MPI's internal reference counting for error handlers.

src/mpi4py/MPI.pyi
dalcinl: Please note you are commenting on a line I'm removed. 

The new signature is
```python
def Get(self, key: str) -> Optional[str]: ...
```
Unfortunately, I believe Python's typing does not allow you to specify size bounds for string.

Also note that the `Info.Get()` method is still there for backward compatibility. I added a new method `Info.Get_string()`, which is implemented with the new MPI-4 routine `MPI_Info_get_string()`.

PR #176, opened 2022-02-23
Add support for scikit-build

dalcinl:


leofang:
Thanks, @dalcinl, this is nice. Can we document how to use scikit-build to build and the env var `MPI_BUILD_BACKEND`?

.github/workflows/ci.yml
```diff
@@ -141,3 +141,66 @@ jobs:
...
+      with:
+        python-version: ${{ matrix.py }}
+
+    - name: Install packaging tools
+      run:  python -m pip install build
+
+    - name: Build sdist
+      run:  python -m build -s
+      env:
+        MPI_BUILD_BACKEND: skbuild
```
leofang, line 180: Shouldn't this be `MPI4PY_BUILD_BACKEND`?

dalcinl: Yes, it should, great catch!. Although to build an sdist, it does not really matter. I'll fix it in an upcoming PR.

dalcinl:
@leofang I still have a few things to do related to packaging in a next PR (for example, stop bundling Cython-generated code in the sdist, and require Cython as a build dependency). 

> Can we document ...

:rofl: :rofl: :rofl: 

I was not sure about advertising this new scikit-build stuff just yet, but perhaps we can add a note that this is still experimental. Of course, you had to note the PR, and the documentation issue unravels once again :fearful: ...

leofang:
Right, it'd be nice to make Cython a build-time dependency. I was under the impression that you wanna minimize the number of dependencies so I didn't bother to suggest 😄

We can definitely say that the skbuild support is experimental and we welcome feedback.

dalcinl:
> I was under the impression that you wanna minimize the number of dependencies

Well, that's always important. But remember mpi4py is old, really old. Back in the days, Cython was new and potentially unstable. We did not have CI services. Python packaging was messier, way messier of what is today. So the recommendation was to ship the C sources in the sdist, just to prevent things breaking. But nowadays, I believe these concerns are no longer valid, or at least not so pressing. If a given Cython version breaks things out of a bug, you can ask the modern build backends to ignore specific versions of a package.

Regarding documentation, it is quite trivial to explain, actually. You just set the environment variable, you run `python -m build` (or `python -m pip wheel`) and you get a wheel file in the `dist/` folder. You install the wheel with whatever installer you use, usually `python -m pip install dist/mpi4py-*.whl`.

PR #179, opened 2022-02-27
Handle byte order markers at the beginning of format strings

### LINKED ISSUE #177: KeyError with mpi4py.MPI.message_basic with explicit little endianness set
This was pretty time-consuming to wrap my head around, so I am posting this issue to hopefully save others some time! 
This issue is related to: https://github.com/mpi4py/mpi4py/issues/81

Thanks for all of your work on this project. =) The lab I am part of uses `mpi4py` for our distributed analyses.

I do not understand enough about the complexities of broadcasting, and computer architectures, to know if handling my issue is something that would be good for the project (i.e. if allowing numpy datastructured with explicit endian-ness set makes sense), but it would have been helpful to add some error-handling there with clearer messaging so that I could immediately understand why my numpy datastructure was failing to broadcast, and adjust things in my code.

Test Case:

```
import sys
import numpy as np
import mpi4py
from mpi4py import MPI


comm = MPI.COMM_WORLD

if comm.rank == 0:
    print("--- Info ---")
    print(f"{mpi4py.__version__=}")
    print(f"{sys.byteorder=}")
    print()
    print("--- Starting Bcast with numpy native endianness set ---")
comm.Barrier()


a = np.arange(10).astype(np.float64)
print(f"{comm.rank=}: {a.dtype=}, {a.dtype.str=}, {a.dtype.byteorder=}")
comm.Bcast(a, root=0)

comm.Barrier()
if comm.rank == 0:
    print("Success")
    print()
    print("--- Starting Bcast with explicit little endianness set ---")
comm.Barrier()

a = np.arange(10).astype(np.float64)
a = a.newbyteorder("<")
print(f"{comm.rank=}: {a.dtype=}, {a.dtype.str=}, {a.dtype.byteorder=}")
comm.Bcast(a, root=0)

comm.Barrier()
if comm.rank == 0:
    print("---")
```

Output:

```
mpirun -np 2 python testb.py
--- Info ---
mpi4py.__version__='3.1.3'
sys.byteorder='little'

--- Starting Bcast with numpy native endianness set ---
comm.rank=0: a.dtype=dtype('float64'), a.dtype.str='<f8', a.dtype.byteorder='='
comm.rank=1: a.dtype=dtype('float64'), a.dtype.str='<f8', a.dtype.byteorder='='
Success

--- Starting Bcast with explicit little endianness set ---
comm.rank=0: a.dtype=dtype('<f8'), a.dtype.str='<f8', a.dtype.byteorder='<'
comm.rank=1: a.dtype=dtype('<f8'), a.dtype.str='<f8', a.dtype.byteorder='<'
  File "/Users/richard/Projects/CHIME/quick/mpi4py_endian/testb.py", line 32, in <module>
  File "/Users/richard/Projects/CHIME/quick/mpi4py_endian/testb.py", line 32, in <module>
    comm.Bcast(a, root=0)
  File "mpi4py/MPI/Comm.pyx", line 690, in mpi4py.MPI.Comm.Bcast
    comm.Bcast(a, root=0)
  File "mpi4py/MPI/Comm.pyx", line 690, in mpi4py.MPI.Comm.Bcast
  File "mpi4py/MPI/msgbuffer.pxi", line 531, in mpi4py.MPI._p_msg_cco.for_bcast
  File "mpi4py/MPI/msgbuffer.pxi", line 534, in mpi4py.MPI._p_msg_cco.for_bcast
  File "mpi4py/MPI/msgbuffer.pxi", line 495, in mpi4py.MPI._p_msg_cco.for_cco_send
  File "mpi4py/MPI/msgbuffer.pxi", line 511, in mpi4py.MPI._p_msg_cco.for_cco_recv
  File "mpi4py/MPI/msgbuffer.pxi", line 203, in mpi4py.MPI.message_simple
  File "mpi4py/MPI/msgbuffer.pxi", line 203, in mpi4py.MPI.message_simple
  File "mpi4py/MPI/msgbuffer.pxi", line 145, in mpi4py.MPI.message_basic
KeyError: '<d'
  File "mpi4py/MPI/msgbuffer.pxi", line 145, in mpi4py.MPI.message_basic
KeyError: '<d'
```

 On every computer we use the native order is little endian, and that is distinguised in the numpy dtype as `=`. However, what seems to happen is that as HDF5 explicitly says that they are little-endian floats, the dtype byte order becomes explicitly little endian (i.e. <) for the array. mpi4py's `TypeDict` does not carry references for endian-ness, and cannot find the numpy dtype among its keys. 
 
The "workaround" is to fiddle the byte order in the array via something like: wdt = wdt.newbyteorder("="), but it is not clear if that is the best approach. The workaround in https://github.com/mpi4py/mpi4py/issues/81 also works.

dalcinl:
Fixes #177.

### DIFF
--- a/src/mpi4py/MPI/ascaibuf.pxi
+++ b/src/mpi4py/MPI/ascaibuf.pxi
@@ -67,6 +67,7 @@ cdef int Py_GetCAIBuffer(object obj, Py_buffer *view, int flags) except -1:
     cdef bint readonly = 0
     cdef Py_ssize_t s, size = 1
     cdef Py_ssize_t itemsize = 1
+    cdef char byteorder = c'|'
     cdef char typekind = c'u'
     cdef bint fixnull = 0
 
@@ -89,6 +90,7 @@ cdef int Py_GetCAIBuffer(object obj, Py_buffer *view, int flags) except -1:
     for s in shape: size *= s
     if dev_ptr is None and size == 0: dev_ptr = 0 # XXX
     buf = PyLong_AsVoidPtr(dev_ptr)
+    byteorder = <char>ord(typestr[0:1])
     typekind = <char>ord(typestr[1:2])
     itemsize = <Py_ssize_t>int(typestr[2:])
 
@@ -97,6 +99,23 @@ cdef int Py_GetCAIBuffer(object obj, Py_buffer *view, int flags) except -1:
             "__cuda_array_interface__: "
             "cannot handle masked arrays"
         )
+    if byteorder == c'<': # little-endian
+        if not is_little_endian():
+            raise BufferError(
+                f"__cuda_array_interface__: "
+                f"typestr {typestr!r} "
+                f"with non-native byte order")
+    elif byteorder == c'>': # little-endian
+        if not is_big_endian():
+            raise BufferError(
+                f"__cuda_array_interface__: "
+                f"typestr {typestr!r} "
+                f"with non-native byte order")
+    elif byteorder != c'|':
+        raise BufferError(
+            f"__cuda_array_interface__: "
+            f"typestr {typestr!r} "
+            f"with unrecognized byte order")
     if size < 0:
         raise BufferError(
             f"__cuda_array_interface__: "
--- a/src/mpi4py/MPI/msgbuffer.pxi
+++ b/src/mpi4py/MPI/msgbuffer.pxi
@@ -81,14 +81,37 @@ cdef inline bint is_IN_PLACE(object obj):
 
 #------------------------------------------------------------------------------
 
+cdef inline const char *getformat(const char format[]) except NULL:
+    cdef char byteorder = format[0]
+    if byteorder == c'@': # native
+        return format + 1
+    if byteorder == c'<': # little-endian
+        if not is_little_endian(): raise ValueError(
+            f"format string {pystr(format)!r} "
+            f"with non-native byte order")
+        return format + 1
+    if byteorder == c'>': # big-endian
+        if not is_big_endian(): raise ValueError(
+            f"format string {pystr(format)!r} "
+            f"with non-native byte order")
+        return format + 1
+    # passthrough
+    return format
+
+cdef inline Datatype lookup_datatype(object key):
+    try:
+        return <Datatype?> TypeDict[key]
+    except KeyError:
+        raise KeyError(f"cannot map {key!r} to MPI datatype")
+
 cdef inline Datatype asdatatype(object datatype):
     if isinstance(datatype, Datatype):
         return <Datatype> datatype
-    return <Datatype?> TypeDict[datatype]
+    return lookup_datatype(datatype)
 
 cdef inline Datatype getdatatype(const char format[]):
     if format == BYTE_FMT: return __BYTE__
-    return <Datatype?> TypeDict[pystr(format)]
+    return lookup_datatype(pystr(getformat(format)))
 
 @cython.final
 @cython.internal
--- a/src/mpi4py/MPI/asbuffer.pxi
+++ b/src/mpi4py/MPI/asbuffer.pxi
@@ -52,6 +52,16 @@ BYTE_FMT[1] = 0
 
 #------------------------------------------------------------------------------
 
+cdef inline int is_big_endian() nogil:
+    cdef int i = 1
+    return (<char*>&i)[0] == 0
+
+cdef inline int is_little_endian() nogil:
+    cdef int i = 1
+    return (<char*>&i)[0] != 0
+
+#------------------------------------------------------------------------------
+
 include "asdlpack.pxi"
 include "ascaibuf.pxi"
 
--- a/test/test_msgspec.py
+++ b/test/test_msgspec.py
@@ -377,6 +377,22 @@ class TestMessageSimpleNumPy(unittest.TestCase,
     def array(self, typecode, initializer):
         return numpy.array(initializer, dtype=typecode)
 
+    def testByteOrder(self):
+        sbuf = numpy.zeros([3], 'i')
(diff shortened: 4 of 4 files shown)


dalcinl:
@anjakefala Please take a look.

src/mpi4py/MPI/msgbuffer.pxi
```diff
@@ -81,14 +81,41 @@ cdef inline bint is_IN_PLACE(object obj):
...
 
+cdef inline int is_bigendian() nogil:
+    cdef int i = 1
+    return (<char*>&i)[0] == 0
+
+cdef inline const char *getformat(const char format[]) except NULL:
+    cdef char byteorder = format[0]
+    if byteorder == c'@': # native
+        return format + 1
+    if byteorder == c'<': # little-endian
```
leofang, line 92:
```suggestion
    elif byteorder == c'<': # little-endian
```

src/mpi4py/MPI/msgbuffer.pxi
```diff
@@ -81,14 +81,41 @@ cdef inline bint is_IN_PLACE(object obj):
...
+        if is_bigendian(): raise ValueError(
+            f"format string {pystr(format)!r} "
+            f"with non-native byte order")
+        return format + 1
+    if byteorder == c'>': # big-endian
```
leofang, line 97:
```suggestion
    elif byteorder == c'>': # big-endian
```

leofang (commented):
LGTM except two nitpicks

src/mpi4py/MPI/msgbuffer.pxi
dalcinl: Why? If any of the `if` branches is followed, the function either returns or raises exception.  So using successive `if` branches vs `if/elif` it is just as a matter of style with not semantic differences. Or am I wrong?

src/mpi4py/MPI/msgbuffer.pxi
dalcinl: Same thing here, just a matter of style.

src/mpi4py/MPI/msgbuffer.pxi
leofang: Sorry, I don't remember what I was thinking earlier. 😞

leofang (approved):

leofang:
Note: I think this patch fixes the buffer protocol and the CAI protocol, but not the DLPack protocol because apparently DLPack forgot about endianness. Raised an issue to track it (https://github.com/dmlc/dlpack/issues/97).

leofang:
> I think this patch fixes the buffer protocol and the CAI protocol

Sorry, I was wrong. It seems strictly speaking CAI could also have an issue, since we currently ignore the endianness there (`typestr[0]`):
https://github.com/mpi4py/mpi4py/blob/84925d924a65a32e1dd34a65b9851b5d3a60e6f1/src/mpi4py/MPI/ascaibuf.pxi#L92-L93
Frankly this is an oversight not just on us but also on those who contributed to CAI (well, me included...) -- CUDA is little endian so it does not make sense to accept/exchange big endianness. What should we do here? No one complained yet so just ignore it (since in practice it doesn't matter)?

dalcinl:
@leofang It would be rather trivial to add a check for native byte order to CAI, although I'm not sure whether it is worth it, DLPack takes precedence over CAI. So far I has been considering CAI as deprecated/legacy stuff. However, errors should never pass silently.

dalcinl:
@leofang I implemented a few minor refactors. Afterwards, I added proper checks in CAI. Please take a new look.

anjakefala:
@dalcinl This is great! Once there is a tar, I happily offer to test it with my data.

src/mpi4py/MPI/ascaibuf.pxi
```diff
@@ -97,6 +99,23 @@ cdef int Py_GetCAIBuffer(object obj, Py_buffer *view, int flags) except -1:
             "__cuda_array_interface__: "
             "cannot handle masked arrays"
         )
+    if byteorder == c'<': # little-endian
+        if not is_little_endian():
+            raise BufferError(
+                f"__cuda_array_interface__: "
+                f"typestr {typestr!r} "
+                f"with non-native byte order")
+    elif byteorder == c'>': # little-endian
```
leofang, line 108: Big endian?

leofang:
> So far I has been considering CAI as deprecated/legacy stuff. However, errors should never pass silently.

That's what I thought too, until after seeing some Python only libraries that can't use dlpack. Numba is one representative example 😄

leofang (approved):

dalcinl:
> until after seeing some Python only libraries that can't use dlpack

So DLPack cannot deal with complex double precision numbers, no support for readonly, and no support for endianness. On top of that the standardized Python protocol cannot be implemented with Python-only code. Looks like issues are piling up...

dalcinl:
> This is great! Thank you so much! Once there is a tar, I happily offer to test it with my data.

I do not have an ETA for a new release tarball. However, installing mpi4py from git is trivial, just do:
```shell
python -m pip install https://github.com/mpi4py/mpi4py/tarball/master
```

I'm going to merge this PR right now. The tests I've added are all passing in our CI. I have not tested on an actual big endian machine yet, but if we ever do, the added tests will catch any issues.

leofang:
> So DLPack cannot deal with complex double precision numbers, no support for readonly, and no support for endianness. On top of that the standardized Python protocol cannot be implemented with Python-only code. Looks like issues are piling up...

Complex numbers can be handled, but yeah the rest is being discussed.

dalcinl:
> Complex numbers can be handled

Sorry, I made I mistake in my comment, I edited it. 

DLPack cannot handle <ins>quadruple</ins> precision complex numbers.

PR #196, opened 2022-04-28
BUG: Fix typing stubs syntax for python 3.6 and 3.7

henrikhierl:


henrikhierl:
Hi @dalcini your suggest is incorporated. To the question if I need it, since the listed supported version of Python includes 3.6/3.7 I would expect all parts should work or at least not break anything.
As currently no mpi4py dependent project can run mypy checks with Python 3.6/3.7, I would fix this bug or change the mpi4py Python requirement to 3.8 as of mpi4py version 3.1.0.

henrikhierl:
I see your point. I'm not an expert, could it be possible not to include the *.pyi and py.typed files for Python versions earlier than 3.8?

dalcinl:
@henrikhierl You raise very valid points. I'm looking for the best way to handle your concerns. So far, I'm thinking about adding a minor fix for Python 3.8 (the signature mismatch I reported above), and exclude the *.pyi and py.typed files as you suggested. Once this work is ready, I'll mention you in the upcoming pull request, we can continue the discussion there.

leofang:
Are we sure we still wanna support Python 3.7? According to NEP-29 it should be dropped by now.
https://numpy.org/neps/nep-0029-deprecation_policy.html

dalcinl:
> Are we sure we still wanna support Python 3.7?

Did you mean Python 3.6? Otherwise, I'm not following. 

In any case, I always to do my best to support legacy stuff, you surely know very well know that HPC facilities are sometimes struck in the past regarding to software.

leofang:
> Did you mean Python 3.6? Otherwise, I'm not following.

According to NEP-29, Python 3.7 support was dropped after `Dec 26, 2021`.

> In any case, I always to do my best to support legacy stuff, you surely know very well know that HPC facilities are sometimes struck in the past regarding to software.

Sure, no objection. Though the missing of typing support isn't a blocker for actually using mpi4py. Just don't turn on your type checker...

PR #206, opened 2022-05-09
mpi4py.futures: Embed remote exception tracebacks as strings

### LINKED ISSUE #204: mpi4py.futures: embed stringification of remote traceback in local traceback
Tracebacks of exceptions raised remotely via Executor.map() don't show the call stack of the worker but only

> .../concurrent/futures/_base.py", line 391, in __get_result
>     raise self._exception

Python's internal concurrent.futures shows the remote traceback through the following "hack" at https://github.com/python/cpython/blob/5bc2390229bbcb4f13359e867fd8a140a1d5496b/Lib/concurrent/futures/process.py#L116

```
# Hack to embed stringification of remote traceback in local traceback

class _RemoteTraceback(Exception):
    def __init__(self, tb):
        self.tb = tb
    def __str__(self):
        return self.tb

class _ExceptionWithTraceback:
    def __init__(self, exc, tb):
        tb = ''.join(format_exception(type(exc), exc, tb))
        self.exc = exc
        # Traceback object needs to be garbage-collected as its frames
        # contain references to all the objects in the exception scope
        self.exc.__traceback__ = None
        self.tb = '\n"""\n%s"""' % tb
    def __reduce__(self):
        return _rebuild_exc, (self.exc, self.tb)

def _rebuild_exc(exc, tb):
    exc.__cause__ = _RemoteTraceback(tb)
    return exc
```

I suggest a similar technique be applied at https://github.com/mpi4py/mpi4py/blob/d56ecfd3ca29ef018420c964359e759fbe230f1c/src/mpi4py/futures/_lib.py#L59

```
def sys_exception():
    exc = sys.exc_info()[1]
    exc.__traceback__ = None
    return exc
```

dalcinl:
Resolves #204


jcphill:
Thanks, that was fast!
I assume you've tested with -n 2 and higher to be check inter-process communication.
I'd suggest "rank X of Y on Z" rather than "worker ..." just because rank 0 is a worker only when size = 1 (right?) and the term rank is unambiguous.
Also, it's possible to interpret "rank 2 of 4 on host23" as meaning there are 4 ranks on host23 rather than 4 ranks total. Clearer would be "executor size 4 exception from rank 2 on host23" or similar.

dalcinl:
> I assume you've tested with -n 2 and higher to check inter-process communication.

Of course! I guess you did not notice the tests I've added. These tests explicitly ensure that remote exceptions are pickled and sent back from workers to the caller and the `__cause__` attribute is properly set. 

> I'd suggest "rank X of Y on Z" rather than "worker ..." just because rank 0 is a worker only when size = 1 (right?) and the term rank is unambiguous. Unless this isn't the rank in COMM_WORLD, in which case it's definitely going to confuse people who don't understand the internals of the library.

I choose this wording on purpose. There are three different ways to mpi4py.futures constructs the MPI process pool: 

1. MPI spawn
2. MPI-1 compatibility mode when MPI spawn is not supported or cannot be used
3. MPICommExecutor

In all these cases, the workers run on an **INTER**communicator which is crated out of the MPI spawning process (case 1), or splitting COMM_WORLD (case 2) or a user-provided comm (case 3) to then create an intercomunicator. 

In case 1, all the workers live on their own COMM_WORLD, and the master process (the one creating the executor instance) live on a different COMM_WORLD. In case 2, all the process live in the same COMM_WORLD, master is rank 0, the workers are the other ranks. Case 3 is similar to 2, but the the user-provided comm may or may not be COMM_WORLD.

All that being said, I hope you understand that referring to "rank or COMM_WORLD" is highly inaccurate or even misleading. IMHO, the best we can do is to say "worker `id` of `N` on `host`", were `N` is the number or workers and `id` is in `[0, N)`. Incidentally, `id` and `N` are the (local) rank and size of the worker side of the intercommunicator.

> Also, it's possible to interpret "rank 2 of 4 on host23" as meaning there are 4 ranks on host23 rather than 4 ranks total.
> Clearer would be "executor size 4 exception from COMM_WORLD rank 2 on host23" or similar, or don't bother with the size and just "executor exception from COMM_WORLD rank 2 on host23"

I'm still open to discuss the exact message, but I hope I convinced you that we should not refer to "rank" and "COMM_WORLD".

What about the following?

```
### worker {id} of a total of {N} workers running on {hostname}
```

dalcinl:
> There are three different ways to mpi4py.futures constructs the MPI process pool:

Actually, there is a fourth way: using MPI accept/connect routines. I bet no one is using that in production.

jcphill:
> Of course! I guess you did not notice the tests I've added.

Sorry. I just saw "-n 1" in the example.

> All that being said, I hope you understand that referring to "rank or COMM_WORLD" is highly inaccurate or even misleading. IMHO, the best we can do is to say "worker `id` of `N` on `host`", were `N` is the number or workers and `id` is in `[0, N)`. Incidentally, `id` and `N` are the (local) rank and size of the worker side of the intercommunicator.

I hadn't considered the spawn case. Calling it "worker" is fine, but then we really do need to print the size so the user notices when it is different than the number of processes on the mpiexec command line.

> What about the following?
> 
> ```
> ### worker {id} of a total of {N} workers running on {hostname}
> ```

This has the same ambiguity since "running on {hostname}" can still be associated with "a total of {N} workers".

Giving the end user more information makes it very clear what happened:

```
### executor with {N} workers caught exception
### raised by worker {id} on {hostname}
```

dalcinl:
> > the best we can do is to say "worker `id` of `N` on `host`"
> 
> Calling it "worker" is fine, but then we really do need to print the size
> 

My original proposal was `### worker {id} of {N} on {host}`. I was printing the size, perhaps you missed it.

> > `### worker {id} of a total of {N} workers running on {hostname}`
> 
> This has the same ambiguity ...

Granted!  Perhaps `### worker {id}, of a total of {N} workers, running on {hostname}` is a bit better.

However, I still believe that my original
```
### Worker {id} of {N} on {host}
```
should be enough. The extra info is printed succinctly, on a single line. Anyone with the bare minimum knowledge of MPI using mpi4py on a cluster or supercomputer should know that processes do not necessarily run on the same host.

src/mpi4py/futures/_lib.py
```diff
@@ -55,13 +56,42 @@ def setup_mpi_threads():
...
+def _wrap_exc(exc, tb):
+    exc.__cause__ = None
+    exc.__context__ = None
     exc.__traceback__ = None
+    return _ExceptionWrapper(exc, tb)
+
+
+def _unwrap_exc(exc, tb):
+    exc.__cause__ = RemoteTraceback(tb)
     return exc
```
leofang, line 77: nit: could you move this definition to the front, after `RemoteTraceback` and before everything else? I got confused at my first glance "where is this defined"...

src/mpi4py/futures/pool.py
```diff
@@ -116,7 +116,7 @@ def submit(self, fn, *args, **kwargs):
             task = (fn, args, kwargs)
             self._pool.push((future, task))
             return future
-    if sys.version_info >= (3, 8):
+    if sys.version_info >= (3, 8):  # pragma: no branch
```
leofang, line 119: What does this pragma do?

dalcinl: It is an annotation for test coverage. I usually run it with latest Python, so the false branch is never taken, then coverage marks a miss. https://app.codecov.io/gh/mpi4py/mpi4py

src/mpi4py/futures/_lib.py
dalcinl: I'm not sure how exactly you want it defined. I'll do it in the following order

* RemoteTraceback
* _unwrap_exc
* _ExceptionWrapper
* _wrap_exc

so there can be no confusion, things are defined before they are used.

PS: This pattern (use something first in a function/method, define it later in the module) is quite common in Python code.

src/mpi4py/futures/pool.py
leofang: Neat!

src/mpi4py/futures/_lib.py
leofang: Yeah it's not uncommon, which was why I said it's a nitpick 🙂

leofang: Thanks.

jcphill:
This is working great in our environment. Thanks again!

Issue #71, opened 2021-07-26
Document more env vars?


leofang:
We should make a distinction between env vars used at build time and those used at runtime. Runtime env vars include:
- `MPI4PY_MAX_WORKERS`
- `MPI4PY_SERVICE`
- `MPI4PY_SERVER_HOST`
- `MPI4PY_SERVER_BIND`
- `MPI4PY_SERVER_PORT`

and build time env vars include:
- `MPECC`
- `VTCC`
- `MPICFG`
- `MPE_LOGFILE_PREFIX`
- `VT_FILE_PREFIX`
- `I_MPI_ROOT`
- `MSMPI_INC`
- `MSMPI_LIB`
- `MSMPI_BIN`
- `CCP_HOME`
- `MPICC`

Some of them are not mpi4py's env vars, but maybe worth mentioning?

leofang:
Maybe we should also accept `MPI4PY_CFG` as an alias of `MPICFG`...

dalcinl:
1. The envvars related to `mpi4py.futures` should be properly namespaced as `MPI4PY_FUTURES_XXX`, and the previous names should be deprecated. Let me look at this first before any attempt to further document them. The really important one is `MAX_WORKERS` and it is already documented in the `mpi4py.futures` section. The other variables are related to relatively obscure and not frequently used MPI dynamic process management I implemented and support just for completeness. I would rather prefer to remain silent for the time being.
2. About the build-time environment vars, I'm not really sure. As time passes, more and more users consume mpi4py from pre-built packages, specially on Windows, so I hardly see the need to document `I_MPI_ROOT`, and `MSMPI_XXX`/`CPP_HOME` variables. `MPICC` is already documented. I would say that `MPE*` (used to be available within MPICH) and `VT*` (used to be available within Open MPI) are legacy stuff at this point. That leaves us with `MPICFG`. This one is not documented at all. You are right, the name sucks. Again, its use is rather specialized. Anyway, we could discuss a better name (e.g. `MPI4PY_MPI_CONFIG`) and add a note in the install section.

dalcinl:
As you can see, I always find good excuses to not write documentation.

PR #218, opened 2022-06-02
Various build/test/docs fixes and enhancements

dalcinl:


dalcinl:
@leofang You may want to update issue #71  after this one is merged.

docs/source/install.rst
```diff
@@ -1,4 +1,98 @@
...
+.. envvar:: MPILD
+
+   The :program:`mpicc` compiler wrapper is also used for linking the
+   :mod:`mpi4py.MPI` C extension module. Alternatively, you can set
+   the :envvar:`MPILD` environment variable to the full path or
+   command corresponding to the MPI-aware C linker.
+
+.. envvar:: MPICFG
+
+   Convenience alias for :envvar:`MPI4PY_BUILD_MPICFG`.
```
leofang, line 51: Should we deprecate it?

docs/source/install.rst
```diff
@@ -1,4 +1,98 @@
...
+
+.. envvar:: MPI4PY_BUILD_MPICFG
+
+   If the MPI implementation do not provide a compiler wrapper, or it
```
leofang, line 55:
```suggestion
   If the MPI implementation does not provide a compiler wrapper, or it
```

test/arrayimpl.py
```diff
@@ -526,19 +512,23 @@ def as_raw(self):
...
+                yield array, typecode
+            else:
+                for prod in itertools.product(*args):
+                    yield (array, typecode) + prod
+    del loop.array
+    del loop.typecode
+
+
+def test(case, **kargs):
+    return case.subTest(
```
leofang, line 530: I am lost, what is `type(case)`?

test/arrayimpl.py
```diff
@@ -526,19 +512,23 @@ def as_raw(self):
...
+        typecode=loop.typecode,
+        backend=loop.array.backend,
```
leofang, line 532: and why would `loop.typecode` and `loop.array` be defined at this point?

test/arrayimpl.py
```diff
@@ -526,19 +512,23 @@ def as_raw(self):
...
 
 
-def subTest(case, skip=(), skiptypecode=()):
+def loop(*args):
     for array in ArrayBackends:
-        if array.backend == skip:
-            continue
-        if array.backend in skip:
-            continue
+        loop.array = array
```
leofang, line 517: Not a big fan of attaching attributes to a function... If `loop` has to have attributes, can't we just turn it into a class that has `__call__`?

docs/source/install.rst
```diff
@@ -1,4 +1,98 @@
...
+   is not installed in a default system location, all relevant build
+   information like include/library locations and library lists can be
+   provided in a ini-style configuration file under a ``[mpi]``
```
leofang, line 58: nit:
```suggestion
   provided in an ini-style configuration file under a ``[mpi]``
```

docs/source/install.rst
```diff
@@ -1,4 +1,98 @@
...
+   section. mpi4py can then be asked to use the custom build
+   information by setting the :envvar:`MPI4PY_BUILD_MPICFG`
+   environment variable to the full path of the configuration file. As
+   an example, see the :file:`mpi.cfg` file located in the top level
+   mpi4py source directory.
+
+.. envvar:: MPI4PY_BUILD_CONFIGURE
```
leofang, line 65: This is handy 👍

docs/source/install.rst
dalcinl: I'm sick to the bones with Python's packaging-related deprecations and I don't want to contribute to this trend. Moreover, the main reason I'm adding the `MPI4PY_BUILD_MPICFG` name is to provide proper namespacing. I'm even considering whether we should support `MPI4PY_BUILD_MPICC` and so on, keeping the previous ones as aliases.

test/arrayimpl.py
dalcinl: Because of the way `loop()` and `test()` are used. The point of these two routines, silly as they may look, it to reduce the verbosity of looping over arraytypes/typecodes and using the subTest context manager at the same time.

test/arrayimpl.py
dalcinl: I use function attributes sparingly, as a way to track and hide things that would otherwise have to be stored in global variables. At this point, I do not want to complicate the implementation of all this silly extra testing support. If mpi4py would use a more advanced testing framework, then all this mess would probably not be there. But I do not want to depend on stuff outside Python's stdlib.

dalcinl:
The testsuite is definitely not nice code. mpi4py testing is not a fun part of maintaining this project. I'm quite sick of it, actually. I'm totally tired of chasing issues related to poor development practices in projects mpi4py depends on but I'm not responsible of. Therefore, for the time being, I do my best to keep the stuff under `test/` working, but as it is not user-facing, I do not care too much about it being good looking.

docs/source/install.rst
leofang: Perhaps I wasn't being clear, I meant deprecate those env vars without a namespace scope. I am in favor of all env vars scoped by the project name to avoid potential issues.

leofang (approved):

docs/source/install.rst
dalcinl: I understood you. But what do you mean by "deprecation"? Emit deprecation warnings? That's what I'm not in favor of: more packaging/build deprecations, I'm sick of it. Without emitting a deprecation warning, the whole discussion is moot. But even if we emit a deprecation warning, I don't see any reason (besides naming aesthetics) for actually remove the short names with the obvious compatibility breakage.

In my view, there is nothing technically wrong with a short name like `MPICC`, it is more convenient for anyone installing manually in the commmand line, for example. The only reason I would like to add `MPI4PY_BUILD_MPICC` is to provide a project-specific name that can be used to avoid conflicts with other projects using the name `MPICC` with different semantics.

So, let the short names survive, and add namespaced names for all these vars. I'll push a new commit shortly.

docs/source/install.rst
```diff
@@ -1,4 +1,115 @@
...
+
+   Some vendor MPI implementations may not provide complete coverage
+   of the MPI standard, or may provide partial features of newer MPI
+   standard versions while advertising support for an older version.
+   Setting the :envvar:`MPI4PY_BUILD_CONFIGURE` environment variable
+   to a non-empty string will trigger the run of exhaustive checks for
+   the availability of all MPI constants, predefined handles, and
+   routines.
+
+The following environment variables are aliases for the ones described
```
dalcinl, line 73: @leofang Do you like it this way?

leofang: Thanks, Lisandro, LGTM.

leofang (approved):

docs/source/install.rst
leofang: Avoiding conflicts means everything to me and I don't really care aesthetics whatsoever. It's extremely painful to debug in days and finally realize some env vars could be accepted in multiple projects but they expect the same env var to be used in slightly different fashion. As a recent example, `NVCC` in some projects can be prefixed with `ccache` but some explicitly reject this idea. The outcome is the build pipeline failed with an obscure error. There's no way a user or package maintainer can figure it out without lengthy investigation.

PR #241, opened 2022-08-10
allow IPv6 sockets in test_dynproc.py testJoin

### LINKED ISSUE #240: testJoin fails on IPV6-only networks
Debian reports that mpi4py fails to build on IPV6-only build daemons, reported in Debian [Bug#1015912](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1015912). A build log is available at 
https://buildd.debian.org/status/fetch.php?pkg=mpi4py&arch=amd64&ver=3.1.3-2%2Bb1&stamp=1658589239&raw=0

The error message is in testJoin (test_dynproc.TestDPM):
```
ERROR: testJoin (test_dynproc.TestDPM)
  File "/<<PKGBUILDDIR>>/test/test_dynproc.py", line 175, in testJoin
    server.bind((host, 0))
socket.gaierror: [Errno -5] No address associated with hostname

Ran 547 tests in 17.493s

FAILED (errors=1, skipped=10)
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
with errorcode 1.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
```

drew-parsons:
Don't assume IPv4 sockets are available in TestDPM.testJoin
(test_dynproc.py). Instead, use the family reported available by
socket.getaddrinfo().

Use INET (IPv4) by default if available, otherwise INET6 (IPv6) or
UNIX sockets. An undocumented socket family will also be used if no
other is available.

Intended to enable testJoin to pass on IPv6-only networks.

Fixes #240.

### DIFF
--- a/test/test_dynproc.py
+++ b/test/test_dynproc.py
@@ -164,14 +164,27 @@ class TestDPM(unittest.TestCase):
         size = MPI.COMM_WORLD.Get_size()
         rank = MPI.COMM_WORLD.Get_rank()
         server = client = address = None
-        # crate server/client sockets
+        host = socket.gethostname()
+        addresses = socket.getaddrinfo(host, None, 0, socket.SOCK_STREAM)
+        address_families = [ a[0] for a in addresses ]
+        # if both INET and INET6 are available, don't assume the order
+        # is the same on both server and client. Select INET if available.
+        if socket.AF_INET in address_families:
+            socket_family = socket.AF_INET
+        elif socket.AF_INET6 in address_families:
+            socket_family = socket.AF_INET6
+        elif address_families:
+            # allow for AF_UNIX (or other families)
+            socket_family = address_families[0]
+        else:
+            self.skipTest("socket")
+        # create server/client sockets
         if rank == 0: # server
-            server = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            host = socket.gethostname()
+            server = socket.socket(socket_family, socket.SOCK_STREAM)
             server.bind((host, 0))
             server.listen(0)
         if rank == 1: # client
-            client = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            client = socket.socket(socket_family, socket.SOCK_STREAM)
         # communicate address
         if rank == 0:
             address = server.getsockname()


test/test_dynproc.py
```diff
@@ -164,14 +164,29 @@ def testJoin(self):
...
+            socket_family = socket.AF_INET
+        elif socket.AF_INET6 in address_families:
+            socket_family = socket.AF_INET6
+        elif socket.AF_UNIX in address_families:
+            socket_family = socket.AF_UNIX
+        elif address_families:
+            # allow for an undocumented family
+            socket_family = address_families[0]
+        else:
+            raise ConnectionError("internet socket not found")
```
dalcinl, line 182: If networking is disabled, I do not want the test to fail, just skip it.
Please use this `raise unittest.SkipTest("socket")`.

test/test_dynproc.py
```diff
@@ -164,14 +164,29 @@ def testJoin(self):
...
+        host = socket.gethostname()
+        addresses = socket.getaddrinfo(host, None)
+        address_families = [ a[0] for a in addresses if a[1]==socket.SOCK_STREAM ]
+        # if both INET and INET6 are available, don't assume the order
+        # is the same on both server and client. Select INET if available.
+        if socket.AF_INET in address_families:
+            socket_family = socket.AF_INET
+        elif socket.AF_INET6 in address_families:
+            socket_family = socket.AF_INET6
+        elif socket.AF_UNIX in address_families:
```
dalcinl, line 176: I'm not comfortable with using `AF_UNIX`. For it to work, it requires the MPI processes to run on the same host. While that would be OK for a CI builder, it is not OK in general.

Alternatively, you could add something like this:

```
single_host = len(set(MPI.COMM_WORLD.allgather(host))) == 1
...

elif socket.AF_UNIX in address_families and single_host:

```

test/test_dynproc.py
```diff
@@ -164,14 +164,29 @@ def testJoin(self):
...
+            socket_family = socket.AF_UNIX
+        elif address_families:
```
dalcinl, line 178: I would remove this branch handling undocumented families.

test/test_dynproc.py
dalcinl: Moreover, I'm not even sure `AF_UNIX` would actually work in the MPIs out there. I believe they should, but we should check for it. Perhaps we should leave this case for future testing enhancements?

test/test_dynproc.py
```diff
@@ -164,14 +164,29 @@ def testJoin(self):
...
+            # allow for an undocumented family
+            socket_family = address_families[0]
+        else:
+            raise ConnectionError("internet socket not found")
+        # create server/client sockets
```
dalcinl, line 183: Shouldn't we check that `socket_family` is the same in different processes?

test/test_dynproc.py
drew-parsons: I ordered it to prefer INET, or INET6 as second choice. So AF_UNIX would only be activated if INET or INET6 are not available.  It's a fallback case.  The question then is, what would you want the test to do if it's running on a UNIX-only network? Similar in a sense to the "undocumented" case I left in the else clause.  In practice I guess it will never happen.

test/test_dynproc.py
drew-parsons: Can be removed, or the AF_UNIX case could be removed, leaving this else clause to catch "other" families.
The motivation for it was in case some funky new internet family gets invented in the future, AF_QUANTUM say.

test/test_dynproc.py
drew-parsons: That was the motivation for explicitly selecting AN_INET as first choice, so that all processes would invoke the same family.  This assumes that the set of families is the same on all processes, but I didn't assume that the order returned by getaddrinfo() would necessarily be same.

If it's a situation where server has AN_INET (with or without AN_INET6), while client has AN_INET6 only, then it will be a problem.  I think it's not constructive (i.e. too complex) to then require server to switch to AN_INET6.  But checking socket_family across processes is also complex, requiring MPI transfer of the value.  The only step to take if it is done would be to raise an error.  But I imagine an error (or at least a timeout) will be raised anyway if the families don't match.

It seems to me the complexity of catching that scenario would make the test over-engineered.  It would be a strange configuration for client and server to be running on different network families. When it does happen then the test would fail anyway.  Is the scenario likely to occur in real installations?

dalcinl: > It seems to me the complexity of catching that scenario would make the test over-engineered

You know, buddy, I'm so sick of dealing with MPI failures that I admittedly overreact.

> Is the scenario likely to occur in real installations?

Probably not, that would sound like be a crazy setup. But never underestimate the ability of people to mess things up.

Let's keep it as it is for now.

test/test_dynproc.py
drew-parsons: Will do.

dalcinl:
@drew-parsons `AF_UNIX` is not available on Windows
https://github.com/mpi4py/mpi4py/runs/7763801705?check_suite_focus=true

drew-parsons:
> `AF_UNIX` is not available on Windows

I guess that's obvious :)  The case wouldn't be in the way though, it just wouldn't occur on Windows installations. We could simplify the cases by removing the AF_UNIX case, relegating its handling to the catch-all else clause.  That would only create a hypothetical potential conflict if the network runs AF_UNIX (without AF_INET) together with the hypothetical "AF_QUANTUM". Maybe it's better to make that simplification, to mark "else" as handling "AF_UNIX or other families"

dalcinl:
> I guess that's obvious :) The case wouldn't be in the way though, it just wouldn't occur on Windows installations.

What do you mean? The case did happen, look at the logs:

```
  File "test\test_dynproc.py", line 176, in testJoin
    elif socket.AF_UNIX in address_families:
AttributeError: 'module' object has no attribute 'AF_UNIX'
```

I have little Windows expertise, but look slike `socket.getaddrinfo(host, None)` is not returning what we expect.

drew-parsons:
> elif socket.AF_UNIX in address_families:
> AttributeError: 'module' object has no attribute 'AF_UNIX'

I see what you mean.  The socket module itself does not have the attribute defined. In that case squashing it down to "else" will catch it (if needed).

test/test_dynproc.py
```diff
@@ -164,14 +164,29 @@ def testJoin(self):
         size = MPI.COMM_WORLD.Get_size()
         rank = MPI.COMM_WORLD.Get_rank()
         server = client = address = None
-        # crate server/client sockets
+        host = socket.gethostname()
+        addresses = socket.getaddrinfo(host, None)
```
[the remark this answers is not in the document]
drew-parsons: Good point, I'll do that.

test/test_dynproc.py
drew-parsons: Just to check, I think that should be not `raise unittest.SkipTest("socket")` but rather
```
else:
    self.SkipTest("socket")
```
(or `self.SkipTest("network socket not found")`.  Do you prefer the shorter version?)

test/test_dynproc.py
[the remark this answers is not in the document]
drew-parsons: Done in commit abc7954

test/test_dynproc.py
dalcinl: Yes, you are definitely right, use the short version. 

BTW, in an upcoming PR, I should fix a few other cases of the longer version in favor of the shorter one.

drew-parsons: ok, I'll push self.skipTest("socket")

test/test_dynproc.py
```diff
@@ -164,14 +164,27 @@ def testJoin(self):
         size = MPI.COMM_WORLD.Get_size()
         rank = MPI.COMM_WORLD.Get_rank()
         server = client = address = None
-        # crate server/client sockets
+        host = socket.gethostname()
+        addresses = socket.getaddrinfo(host, None, type=socket.SOCK_STREAM)
```
dalcinl, line 168: On Python 2, `getaddrinfo()` does not accept keyword arguments.

```suggestion
        addresses = socket.getaddrinfo(host, None, 0, socket.SOCK_STREAM)
```

dalcinl:
@drew-parsons 

* Please squash all commits under your authorship. I believe GitHub's *Squash and merge* screws authorship. 
* After squashing, test the final patch in Debian. Once you confirm all is good, I'll merge this PR.

PS: I'm not totally happy with a few details, but we had enough with this one. I'll probably work on these issues once I merge all this into master. While mpi4py is a semi-decent test suite for MPI implementations, the ultimate goal is to test mpi4py itself. So I have no shame on skipping tests when I cannot be sure things are reliable.

drew-parsons:
I've rebased to a single commit, acknowledged your getaddrinfo fix in the comment.

drew-parsons:
Thanks @dalcinl .  The Debian builds look fine at https://buildd.debian.org/status/package.php?p=mpi4py

We won't get absolute final confirmation straight away about the reported IPv6 problem since only some of the buildds are IPv6-only.  I think the problematic one is x86-conova-01, since it had the recent fails at https://buildd.debian.org/status/logs.php?pkg=mpi4py&arch=amd64  .  x86-conova-01 should be able to build successfully now when it gets its turn again in the future

PR #247, opened 2022-08-29
Update GitHub Actions

dalcinl:


dalcinl:
@jjhursey I would appreciate your input here.

I'm trying to update my CI infrastructure to use the Ubuntu  22.04 runner images in GitHub Actions. I'm installing Open MPI 4.1.2 with the apt package manager. You can see the `ompi_info` output [here](https://github.com/mpi4py/mpi4py/runs/8065635568?check_suite_focus=true#step:4:271).

1. I had to disable all tests related to `MPI_Comm_spawn_multiple`, there were failing with `MPI_ERR_UNKNOWN` (full logs [here](https://github.com/mpi4py/mpi4py/runs/8064979545?check_suite_focus=true#step:11:1692)). Tests using `MPI_Comm_spawn` were fine, though.
2. After disabling these basic tests, other stuff that depends on `MPI_Comm_spawn` support also failed (deadlock until reaching job timeout), and I got the following output (full logs [here](https://github.com/mpi4py/mpi4py/runs/8065635138?check_suite_focus=true#step:13:23)):
```
A request has timed out and will therefore fail:

  Operation:  LOOKUP: ../../../orte/orted/pmix/pmix_server_pub.c:345

Your job may terminate as a result of this problem. You may want to
adjust the MCA parameter pmix_server_max_wait and try again. If this
occurred during a connect/accept operation, you can adjust that time
using the pmix_base_exchange_timeout parameter.
```

Do you have any advice? Are you aware of bugs related to 4.1.2 that would explain these failures? I know 4.1.2 is not the latest patch-release of the v4.1.x series, but that's what users get from [Ubuntu Jammy](https://packages.ubuntu.com/jammy/libopenmpi3).
If you cannot help me or it would take too much of your time, do not worry. I'll just disable these tests as known failures with 4.1.2 on GitHub actions and move on.

@drew-parsons Just FYI, I'm not sure this is relevant to you.

jjhursey:
@dalcinl I could not reproduce any known issue with spawn using the `master` branch of mpi4py. I used a container to create the build/run environment. I posted the Dockerfile, build, and test output in the Gist below:
 * https://gist.github.com/jjhursey/3f3fd0ea0c394c32cf42e9492bdf1556

However, If I used this branch I was able to see the test case errors ([See this output file](https://gist.github.com/jjhursey/3f3fd0ea0c394c32cf42e9492bdf1556#file-test-error-with-branch-txt)). I'm looking at the tests now to see how I can help.

jjhursey:
I fixed the environment configuration as noted in [this comment](https://gist.github.com/jjhursey/3f3fd0ea0c394c32cf42e9492bdf1556?permalink_comment_id=4285795#gistcomment-4285795)

I was able to run the test case clean.

So I decided to try it in a loop to see if this is an intermittent error. The test case failed after a few successful iterations. It would not always fail in the same test. I saw it fail in `testNoArgs (test_spawn.TestSpawnMultipleSelf)`, `testNoArgs (test_spawn.TestSpawnMultipleWorld)`, `testNoArgs (test_spawn.TestSpawnMultipleWorldMany)`. It did seem to impact `testNoArgs` in my testing, but I don't know if that is just the lack of samples at this point.

```bash
#!/bin/bash

for i in {1..10} ; do
    echo "================================"
    echo "================================ Iteration $i"
    echo "================================"
    timeout -k 33 30 mpiexec -n 1 python test/runtests.py -v -i test_spawn
    RTN=$?
    if [[ $RTN != 0 ]] ; then
        echo "Error: Failed with $RTN"
        exit 1
    fi
    echo ""
done

exit 0
```

drew-parsons:
Thanks for looking into this.  In the Debian tests we've been patching it out with `@unittest.skipMPI('openmpi(<5.0.0)')`, but it will be good when we can stop doing that.

jjhursey:
I spun up a few builds/tests to run in the background while I was working on other things today.

I ran 3 different versions of Open MPI against the Open MPI internal PMIx and the PMIx shipped with Ubuntu 22.04 as an external PMIx. I ran the Python `mpiexec -n 1 python test/runtests.py -v -i test_spawn` in a loop 50 times in a row to look for hangs. Below is what I found.

Status   | Open MPI Release  | OpenPMIx (internal) | OpenPMIx (from Ubuntu 22.04) |
-------- | ------ | ------------------- | ------------------- |
PASS 🎊     | v4.1.2 | 3.2.3           | -- |
PASS 🎊     | v4.1.3 | 3.2.3           | -- |
PASS 🎊    | v4.1.4 | 3.2.3           | -- |
**FAIL** 🔥  | v4.1.2 | --              | 4.1.2 |
**FAIL** 🔥 | v4.1.3 | --              | 4.1.2 |
**FAIL** 🔥 | v4.1.4 | --              | 4.1.2 |

The Ubuntu 22.04 version of OpenPMix was installed with `apt install -y libpmix-dev libpmix2 libevent-dev hwloc`

So it looks like The Open MPI with the internal OpenPMIx 3.2.3 works fine. But with a higher version of OpenPMIx it may result in intermittent hangs. I've not tested beyond this to see if OpenPMix `master` or a newer release in the OpenPMIx 4.x series is impacted. I'll do some more testing in the background and followup with the OpenPMIx / Open MPI communities. @rhc54 FYI.

rhc54:
Huh - that seems weird. Is it possible that OMPI wasn't using the external libevent/hwloc that the PMIx installation was using? I know I've seen segfaults in the past when the MPI layer was using one libevent and PMIx was using another.

I also just heard a case today where OMPI main was inadvertently picking up an external PMIx install (i.e., no configure pointing it in either internal or external directions), yet using everything else internal. Might be the same thing here.

dalcinl:
> The Ubuntu 22.04 version of OpenPMix was installed with `apt install -y libpmix-dev libpmix2 libevent-dev hwloc`

After reading Ralph's comment... Don't you need `libhwloc-dev` as well?

jjhursey:
As a final follow-up here, I ran a testing matrix over the spawn hang issue and filed a ticket with OpenPMIx:
 * https://github.com/openpmix/openpmix/issues/2735

dalcinl:
Many thank, @jjhursey!

PR #253, opened 2022-09-05
mpi4py.futures: Support environment variable `MPI4PY_FUTURES_BACKOFF`

dalcinl:


docs/source/mpi4py.futures.rst
```diff
@@ -133,6 +133,15 @@ only) thread until they are signaled for completion.
      benefit applications dealing with large buffer-like objects like NumPy
      arrays. See :mod:`mpi4py.util.pkl5` for additional information.
 
+   * *backoff*: :class:`float` value specifying the maximum number of seconds a
```
dalcinl, line 136: @mrogowski Is this explanation good enough?

mrogowski: Looks good! You could add something like "specifying 0 will result in an MPI-like busy wait behavior" to further clarify things.

dalcinl: It's complicated. If you set backoff=0, you actually call `sleep(0)` which is IIRC roughly equivalent to `sched_yield()` on Linux. Look at the following behavior: 

```python
In [1]: from time import sleep

In [2]: from os import sched_yield

In [3]: dummy = lambda x: None

In [4]: %timeit dummy(None)
45 ns ± 0.0388 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [5]: %timeit sched_yield()
293 ns ± 0.361 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [6]: %timeit sleep(0)
310 ns ± 0.319 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [7]: %timeit sleep(1e-6)
55.6 µs ± 0.354 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

So it is not really true that backout=0 is equivalent to busy-waiting. In fact, `sleep(0)` forces (or at least encourages?) a context switch, and that's something that busy-waiting prevents (at least until the kernel preempts the running thread after exhaustion of the thread's time slice).

PR #265, opened 2022-10-11
test: Modernize testsuite

dalcinl:
* Use proper unittest.TestCase.assertXXX() methods
* Remove unused import and set-but-unused variables
* Use f-strings and remove .format() / %-operator


leofang:
Just curious, for tedious textual changes like this, do you do it all by hand or with regex (or some sort of automation)?

dalcinl:
I did it by hand :disappointed:. The testsuite code quality is not something I pay the uttermost attention, but this was a much needed cleanup of cruft accumulated over more than 15 years.

PR #292, opened 2023-01-09
docs: Add citation for mpi4py.futures paper in IEEE TPDS

dalcinl:


leofang:
Congrats! I remember you mentioned this work long time ago, glad to see it finally came out! (And hello @mrogowski!)

mrogowski (approved):

leofang:
How about also updating `README.md` and this page https://mpi4py.readthedocs.io/en/stable/citing.html?

dalcinl:
About updating `README.md`, I'm not really sure, but an extra entry would not hurt. One minor conceptual issue is that the paper is specifically about `mpi4py.futures`, and not `mpi4py` as a whole. I would like to prevent confusion or look deceptive to trick folks cite the new mpi4py.futures paper (which is in Marcin's and my convenience) rather than the older mpi4py paper (which is in your convenience). For bad/deceiving citation practices we have enough with the glorified peer-review system, don't we?

Or.. if we do want to cite the two most recent papers, we should explain in which cases each citation would be more appropriate, and not just add it as an extra bullet point. Does it sound reasonable to you guys?


About the "stable" version of docs in RTD, I cannot update that without publishing a new release. Next release should be major 4.0.0 and include the new MPI-4 support in master. I've been waiting for Open MPI to release 5.0.0, but perhaps I should stop waiting and move on.

dalcinl:
@leofang sorry, I forgot to update you. Done in 0126d64c6d27601f0dffe79ab309a6870b5f80e0.

PR #295, opened 2023-01-13
Ensure unlock is called during CAS test

lrbison:
See https://github.com/open-mpi/ompi/issues/11250

This does not solve the root issue as to why the Compare-And-Swap operation is failing, but it allows tests to continue by calling unlock, otherwise the next test is run while the window is still Locked and deadlock occurs.


dalcinl:
As the lock/unlock pattern is used multiple times in these RMA tests, I'm working on a better solution.

PR #304, opened 2023-01-26
build: Fix editable installs (mpi.cfg file)

### LINKED ISSUE #302: `mpi4py.get_config()` fails with in-place builds
Not sure if this is expected:
```python
>>> import mpi4py                                                                                                                              
>>> mpi4py.get_config()                                                                                                                        
  File "/home/leof/miniforge3/envs/cccccc/lib/python3.9/configparser.py", line 846, in items                                                   
    d.update(self._sections[section])                                                                                                          
KeyError: 'mpi'                                                                                                                                
                                                                                                                                               
During handling of the above exception, another exception occurred:                                                                            
                                                                                                                                               
  File "<stdin>", line 1, in <module>                                                                                                          
  File "/home/leof/dev/mpi4py/src/mpi4py/__init__.py", line 107, in get_config                                                                 
    return dict(parser.items('mpi'))                                                                                                           
  File "/home/leof/miniforge3/envs/cccccc/lib/python3.9/configparser.py", line 849, in items                                                   
    raise NoSectionError(section)                                                                                                              
configparser.NoSectionError: No section: 'mpi'
```
The reason is this line
https://github.com/mpi4py/mpi4py/blob/0438892fefd5a5f8361ef7e4b5d4780b0f9db3e2/src/mpi4py/__init__.py#L103
assumes that `mpi.cfg` is placed at the installed dir root, but at least for in-place builds it's not the case.

dalcinl:
Fixes #302.

### DIFF
--- a/conf/mpidistutils.py
+++ b/conf/mpidistutils.py
@@ -944,19 +944,31 @@ class build_ext(cmd_build_ext.build_ext):
             if not self.dry_run:
                 self.config.dump(filename=mpi_cfg)
 
+    def copy_extensions_to_source(self):
+        build_py = self.get_finalized_command('build_py')
+        cmd_build_ext.build_ext.copy_extensions_to_source(self)
+        for ext in self.extensions:
+            if ext.name == 'mpi4py.MPI':
+                fullname = self.get_ext_fullname(ext.name)
+                filename = self.get_ext_filename(fullname)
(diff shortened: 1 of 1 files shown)


dalcinl:
@leofang If you have a recent enough setuptools, try `pip install -e .` and next double check `mpi4py.get_config()` works as espected.

PR #307, opened 2023-02-07
Add support for ULFM

bosilca:
This PR adds support for the ULFM extensions that are slowly getting into the 5.0 MPI standard, and are currently supported by some versions of OMPI and MPICH. This PR checks if the MPIX functions exists, and if not will replace them with stubs that raise an exception.

Signed-off-by: George Bosilca <bosilca@icl.utk.edu>


dalcinl:
I've been reluctant to add support non-standard MPI extensions to mpi4py. However, as the ULFM extensions are supported in both Open MPI and MPICH, I can make an exception with this one. That being said, this PR will need some extra work on the compatibility side, and obviously, also implement support for MPICH. There are also some modifications to files that are git-tracked but autogenerated, and of course there is no way you would figured that out (I know it is bad practice, but to my defense, Python packaging is not fun to mess with to add file generation steps).

Can you point me out to some C/C++/Fortran tests for ULFM? Adding some tests in mpi4py will be the last step to get this feature in.

@jeffhammond Just FYI. We may need to consider this extensions for your ABI work.

bosilca:
Good news, most of these functions have been voted for inclusion in MPI 5.0, so mpi4py can claim the early adopter badge.

I am working on adding basic tests, and I plan to add python support to our own [test suite](https://github.com/ICLDisco/ulfm-testing). I just haven’t figured out how to make the tests optional (depending on the ULFM support being or not enabled in the underlying MPI library). Any suggestions?

Regarding the generated files I’ll be happy to address any concerns, if you can point me to the right way of adding the support.  Or, feel free toi push into my fork if you think it would be simpler.

dalcinl:
> Good news, most of these functions have been voted for inclusion in MPI 5.0

I'm a bit confused. Looks like that the current [Open MPI API](https://github.com/open-mpi/ompi/blob/main/ompi/mpiext/ftmpi/c/mpiext_ftmpi_c.h) does not match the latest API as 2nd-voted by the MPI Forum (eg. see [get_failed](https://github.com/mpi-forum/mpi-standard/blob/a415719349d1ca7d2a15bbd30df22a15acb0194a/chap-ft/ft.tex#L319)/[ack_failed](https://github.com/mpi-forum/mpi-standard/blob/a415719349d1ca7d2a15bbd30df22a15acb0194a/chap-ft/ft.tex#L346) from mpi-forum/mpi-standard#665, see also mpiwg-ft/ft-issues#15).

@bosilca Of course we can continue working on this until both Open MPI and MPICH get in-sync with the latest ULFM API, but I'm not comfortable with merging into master support for an already outdated extension API. 

@hzhou Any estimate how much it would take for you guys to update to the latest ULFM API?

bosilca:
Right, one if the API currently available in the MPI implementations and used by many users, and the other is the API voted by the standard for the MPI 5.0. What is here is what can be used now to write resilient applications in python, under the assumption that until MPI 5.0 is out this API might change. Our users have lived with this constraint for years.

dalcinl:
@bosilca Can I force-push to your branch to update this PR? Otherwise, take a look to branch `feature/MPI-5` in this repo. Things are mostly ready, but I may still force-push to the branch. I took a more lengthy and convoluted approach, but this way will hopefully minimize work once MPI-5.0 is publicly available.

Testing is rather minimal, I'm not proud of it, but it is just enough to test the wrapping. There are things I still don't understand from the new MPI standard chapter. That being said, I run with surprising/unexpected behavior. For example, after comm revocation, a self-`Send()` call would fail as expecteds with `ERR_REVOKED`, but a self-`Isend()` would not. Also, I'm not sure what's the interaction between revocation and failure acknowledgement. It could be just my ignorance, or actual Open MPI issues.

@hzhou The most notable omissions from MPICH respect to current Open MPI are `MPI_Comm_is_revoked()` and `MPI_Comm_iagree()`. Maybe these are rather easy to add?

@hzhou @bosilca Maybe you guys could add `MPIX_Comm_get_failed()` and `MPIX_Comm_ack_failed()` (but keep the old `MPIX_Comm_failure_xxx()` APIs) ? 
AFAICT, these routines already passed final vote in mpi-forum/mpi-issues#581. 
`MPIX_Comm_get_failed()` is just a rename of `MPIX_Comm_failure_get_acked()`.
`MPIX_Comm_ack_failed()` may require some actual coding work to track the number of acked failures.

bosilca:
@dalcinl you should be able to force-push to my branch.

The non-blocking MPI calls are not expected to return an error at invocation time, but they need to be completed with an error. In your example, the `Isend-self` will be posted by waiting on the request will raise the ERR-REVOKE error.

Without passing `--with-ft mpi` to mpirun you will only get a part of the support, basically errors are locally reported, but then it falls back to the normal behavior (aka. abort).

`MPI_Comm_is_revoked()` and `MPI_Comm_iagree()` are extensions we are currently working on (there is also an `MPI_Comm_ispawn` in the pipeline). We can leave them out of this PR, or have support only in OMPI.

While `MPIX_Comm_get_failed()` can easily be implemented, `MPIX_Comm_ack_failed()` is more complicated. In fact swapping the logic from `ack/get` to `get/ack` requires quite some changes in the MPI library. Anyway, as you noticed they passed the vote, so we will implement them.

bosilca:
I saw your push to support error handlers in python, and honestly python has better mechanisms to handle errors. Using this branch and try/catch one can implement working and readable resilient application in few lines of python. 

Here is a small resilient app, that can handle few spares.

```python
from mpi4py import MPI
import os, signal, argparse

parser = argparse.ArgumentParser(description='Resilient manager of spares')
parser.add_argument('-v', '--verbose', action='store_true')
parser.add_argument('spares', metavar='N', type=int, default=0,
                    help='the number of initial spare processes')
args = parser.parse_args()

management_comm = MPI.COMM_WORLD.Dup()
orig_rank = management_comm.rank
active_processes = management_comm.size - args.spares

while True:
  am_i_a_spare = True if management_comm.rank > active_processes else False
  comm = management_comm.Split(am_i_a_spare, 0)

  management_comm.Set_errhandler(MPI.ERRORS_RETURN)
  comm.Set_errhandler(MPI.ERRORS_RETURN)

  if( 0 == management_comm.rank ):
    print("Rank 0 bailing out (original rank ", orig_rank, ")\n" )
    # Don't call exit() because it will call MPI_Finalize and lead to a deadlock
    os.kill(os.getpid(), signal.SIGKILL)

  if am_i_a_spare:
    comm.Free()  # free the spare comm
    try:
      management_comm.barrier()
      print("Completed. spares are leaving ")
      exit(0)
    except MPI.Exception as inst:
      print("Caught exception ", type(inst), inst.args, inst)
      management_comm.Revoke()
  else:  # Not a spare
    # This is the main loop for the active processors
    try:
      print("Before the barrier (rank = ", comm.rank, ")\n")
      comm.Barrier()
    except MPI.Exception as inst:
      print("Caught exception ", type(inst), inst.args, inst)
      comm.Revoke()
      comm.Free()
      management_comm.Revoke()

  print("Before shrink (rank = ", orig_rank, ")\n")
  management_comm = management_comm.Shrink()
```

src/mpi4py/MPI/Comm.pyx
```diff
@@ -232,6 +232,76 @@ cdef class Comm:
...
+
+    def Get_failed(self) -> Group:
+        """
+        Extract the group of failed and acknowledged processes
+        """
+        cdef Group group = Group.__new__(Group)
+        with nogil: CHKERR( MPI_Comm_get_failed(self.ob_mpi, &group.ob_mpi) )
+        return group
+
+    def Ack_failed(self, int num_to_ack: int = -1) -> int:
```
dalcinl, line 260: @bosilca I have a question about my implementation of this method.

Would the following code with the current MPIX API
```c
MPIX_Comm_failure_ack(comm);
```
be semantically equivalent to the following code with the MPI-5.0 API?
```c
MPI_Comm_size(comm, &num_to_ack);
MPI_Comm_ack_failed(comm, num_to_ack, &num_acked);
```

dalcinl:
@bosilca I just force-pushed your branch. I would prefer to merge first this PR adding the core feature and address the addition of examples in an upcoming PR. Some of the Python exception handling can be further simplified using context managers (`with` statement), but I need to think about it. After looking to some of tutorias in SC, I even envision some utility module (e.g. `mpi4py.util.ulfm`) with reusable patterns that would simplify the development of fault tolerant apps in Python.

PS: It is quite rare that hard core MPI developers would get interested in Python+MPI. Out of curiosity, may I ask what drove your contribution?

src/mpi4py/MPI/Comm.pyx
bosilca: > Some of the Python exception handling can be further simplified using context managers (`with` statement), but I need to think about it. After looking to some of tutorias in SC, I even envision some utility module (e.g. `mpi4py.util.ulfm`) with reusable patterns that would simplify the development of fault tolerant apps in Python.

Cool idea, I would love to see such an utility module with reusable patterns. Let me know if I can help.

> PS: It is quite rare that hard core MPI developers would get interested in Python+MPI. Out of curiosity, may I ask what drove your contribution?

Maybe I'm not such a hardcore MPI developer. I'm using python regularly (not to say almost daily), and it is a wonderful and versatile tool for quick tasks/prototyping. My contribution here was motivated by the fact that I believe that MPI lacks critical capabilities for outside-academia uses, and part of this is the lack of support for fault management. So, in order to make MPI more palatable to these communities you need to address their needs (fault management) and speak their languages (not C).

bosilca: Yes, these two cases lead to similar results. To confirm your understanding, this use case is well documented in the `Advice to users` for the `MPI_Comm_ack_failed` function.

We can also implement `MPIX_Comm_get_acked` with the new API by keeping track of the most recent `num_acked` returned from `MPI_Comm_ack_failed` and then extracting the subgroup with that size from the group returned by `MPI_Comm_get_failed`.

src/lib-mpi/mpiulfm.h
```diff
@@ -0,0 +1,145 @@
...
+  return MPIX_Comm_failure_get_acked(comm, group);
+}
+#undef  MPI_Comm_get_failed
+#define MPI_Comm_get_failed PyMPI_Comm_get_failed
+#endif
+#endif
+
+#ifndef PyMPI_HAVE_MPI_Comm_ack_failed
+#ifdef  PyMPI_HAVE_MPIX_Comm_failure_ack
+static int PyMPI_Comm_ack_failed(MPI_Comm comm, int num_to_ack, int *num_acked)
```
dalcinl, line 94: @bosilca I did not quite understand the second paragraph of your comment [here](https://github.com/mpi4py/mpi4py/pull/307#discussion_r1107528400). Please take a look at this piece of code, where I attempt to implement the MPI-5.0 API  with the MPIX API currently available in Open MPI.

This is my implementation of MPI-5.0 `MPI_Comm_ack_failed()` using `MPIX_Comm_failure_ack()` that Open MPI currently provides. As you can see, I check the value of `num_to_ack` on input, and if it is less than the communicator size, I error (because otherwise I cannot satisfy the "ack all failures" semantics). However, note that I'm returning `MPI_UNDEFINED` for `num_acked`, because I don't know how to get the value.  Perhaps I should use `MPIX_Comm_failure_get_acked(comm, &group)` to query the failed group size and set `num_acked`?

The advise to users also say *One may query, without side effect, for the number of currently aknowledged [<- typo is not mine!] process failures in comm by supplying 0 in num_to_ack.*. I'm not sure whether I can support this semantics by querying the failed group with  `MPIX_Comm_failure_get_acked`. In any case, the best we could do is to return num_acked with either zero or failed_group.size.

dalcinl:
> My contribution here was motivated by the fact that I believe that MPI lacks critical capabilities for outside-academia uses, and part of this is the lack of support for fault management. 

That was my guess. I'm 100% on your side.

> So, in order to make MPI more palatable to these communities you need to address their needs (fault management) and speak their languages (not C).

Python was fun, then it became mainstream! :rofl:

dalcinl:
@bosilca Could you please comment in the following implementation of MPI-5.0 `MPI_Comm_ack_failed()` using currently `MPIX_Comm_failure_ack()` and `MPIX_Comm_failure_get_acked()` ? To ease your review, I'm not checking return error codes from MPI calls.
Note the following:

* If `num_to_ack`  is negative or less than comm.size, then error.
* If `num_to_ack` is `0` (zero), then `MPIX_Comm_failure_ack()` is not called, so no side-effects, right?
* If `num_to_ack` is `comm.size` or larger, then `MPIX_Comm_failure_ack()` is called.
* `num_acked` is returned with the size of the failed group.

```c
static int MPI_Comm_ack_failed(MPI_Comm comm, int num_to_ack, int *num_acked)
{
  int size;
  MPI_Group group;
  
  MPI_Comm_size(comm, &size);

  if ((num_to_ack != 0 && num_to_ack < size) || num_acked == NULL) {
    MPI_Comm_call_errhandler(comm, MPI_ERR_ARG);
    return MPI_ERR_ARG;
  }
  
  if (num_to_ack > 0) {
    MPIX_Comm_failure_ack(comm);
  }
  
  MPIX_Comm_failure_get_acked(comm, &group);
  MPI_Group_size(group, num_acked);
  MPI_Group_free(&group);

  return MPI_SUCCESS;
}
```

Can you confirm whether this is this the closet  MPI-5.0 semantics we can support with the current MPIX APIs available in Open MPI?

Additional concerns:
* The MPI-5.0 proposal says nothing about `num_to_ack` < `0`. I assume the sensible thing to do is error, so did I.
* The MPI-5.0 proposal says nothing about `num_to_ack` > `comm.size`. In this case, I'm not erroring, but maybe I should?

dalcinl:
@bosilca I force pushed again. I've reimplemented `MPI_Comm_ack_failed()` as per the comment above, we can change things if you point any issue I missed. BTW, note that the topmost commit is "Unverified", maybe you want to amend it to refresh your signature.

src/lib-mpi/mpiulfm.h
bosilca: Your implementation of PyMPI_Comm_ack_failed seems correct to me. However, your implementation of PyMPI_Comm_get_failed is not, because `MPIX_Comm_failure_get_acked` only returns the errors you have already acknowledged, when the corresponding MPI function is able to return more failures, even those that have not yet been acknowledged. That being said I don't think you can implement the new API using the old one, and have the same expressivity. 

Btw, https://github.com/open-mpi/ompi/pull/10507 is in the final stages and will bring the new ULFM API to OMPI.

My second comment was about implementing the old API with the new API, not sure how relevant it is in this discussion.

dalcinl: > However, your implementation of PyMPI_Comm_get_failed is not, because MPIX_Comm_failure_get_acked only returns the errors you have already acknowledged, when the corresponding MPI function is able to return more failures, even those that have not yet been acknowledged.

Oh, I totally missed that part... You see, I'm still a bit lost following together the original ULFM proposal and the MPI accepted chapter...

Perhaps I should do it the following way?
```c
static int PyMPI_Comm_get_failed(MPI_Comm comm, MPI_Group *group)
{
  MPIX_Comm_failure_ack(comm);
  return MPIX_Comm_failure_get_acked(comm, group);
}
```
What do you think?

> Btw, open-mpi/ompi#10507 is in the final stages and will bring the new ULFM API to OMPI.

What's the expected timeline for that PR to get to v5.0.x? Perhaps I should wait until it is merged to get this one in? In that case, I'll have keep updating your branch, and then ask a final time for to refresh your commit signature, sorry for the noise.
PS: Will the new ULFM API be backported to v4.0.x ?

bosilca: If you do `failure_ack + get_acked` then you automatically ack failed processes on behalf of the user. The bottom line is that the 2 API are not equivalent, it is difficult to implement the new one using the old one. At least not in an easy way. 

The hardway is by having a duplicate communicator that is not visible to the user, and doing what you propose above on the duplicate. This way you can gather information about new failures, but without disturbing the user visible state of the communicator.

The new API PR should be merged later today, and then moved to the next 5.0 rc. The community OMPI 4.0 does not contains ULFM, and I'm not sure we want to backport the new API into our own fork.

src/mpi4py/MPI/Comm.pyx
```diff
@@ -232,6 +232,92 @@ cdef class Comm:
...
+        """
+        cdef type cls = Comm
+        if   isinstance(self, Intracomm): cls = Intracomm
+        elif isinstance(self, Intercomm): cls = Intercomm
+        cdef Comm comm = <Comm>cls.__new__(cls)
+        with nogil: CHKERR( MPI_Comm_shrink(self.ob_mpi, &comm.ob_mpi) )
+        comm_set_eh(comm.ob_mpi)
+        return comm
+
+    # Legacy ULFM interface
```
dalcinl, line 308: @bosilca The legacy API is exposed undocumented, deprecated (warnings raised), and private (as per usual Python naming conventions using initial underscore).

dalcinl:
@bosilca The two top commits are under your name. Could you please sign them and force push your branch? Afterwards, this is ready to merge, no need to wait for open-mpi/ompi#10507 (thanks to the processor guards for the new APIs).

dalcinl:
I'm thinking about `MPI_Comm_ishrink`. Should we add it to this PR, anticipating for the merge of open-mpi/ompi#9925?

bosilca:
I signed my commits, but let's hold this merge until the ishrink is merged (I'll be doing it this afternoon).

dalcinl:
@bosilca @abouteiller I believe something is not right on the Open MPI side with open-mpi/ompi#9925. 

This is with current `open-mpi/ompi@main` and `mpi4py/mpi4py@feature/MPI-5`. 

Please note that my test for `MPI_Comm_ishrink()` does not exercise any actual process failure, so I would expect it to behave just like `MPI_Comm_idup()`.

Running in singleton mode or `mpiexec` with one MPI process is OK.
```
$ mpiexec -n 1 python test/main.py -i test_ulfm -v -k '*World*IShrink'
[0@localhost] Python 3.11.1 (/usr/bin/python)
[0@localhost] numpy 1.24.1 (/home/dalcinl/.local/lib/python3.11/site-packages/numpy)
[0@localhost] MPI 3.1 (Open MPI 5.1.0)
[0@localhost] mpi4py 4.0.0.dev0 (/home/dalcinl/Devel/mpi4py/build/lib.linux-x86_64-cpython-311/mpi4py)
testIShrink (test_ulfm.TestULFMWorld.testIShrink) ... ok

Ran 1 test in 0.000s

OK
```

However, with two or more MPI processes, I'm getting a segfault on waiting for the request I get from ishrink.
```
$ mpiexec -n 2 python test/main.py -i test_ulfm -v -k '*World*IShrink'
[0@localhost] Python 3.11.1 (/usr/bin/python)
[1@localhost] Python 3.11.1 (/usr/bin/python)
[1@localhost] numpy 1.24.1 (/home/dalcinl/.local/lib/python3.11/site-packages/numpy)
[0@localhost] numpy 1.24.1 (/home/dalcinl/.local/lib/python3.11/site-packages/numpy)
[1@localhost] MPI 3.1 (Open MPI 5.1.0)
[1@localhost] mpi4py 4.0.0.dev0 (/home/dalcinl/Devel/mpi4py/build/lib.linux-x86_64-cpython-311/mpi4py)
[0@localhost] MPI 3.1 (Open MPI 5.1.0)
[0@localhost] mpi4py 4.0.0.dev0 (/home/dalcinl/Devel/mpi4py/build/lib.linux-x86_64-cpython-311/mpi4py)
testIShrink (test_ulfm.TestULFMWorld.testIShrink) ... testIShrink (test_ulfm.TestULFMWorld.testIShrink) ... [localhost:388268] *** Process received signal ***
[localhost:388268] Signal: Segmentation fault (11)
[localhost:388268] Signal code: Address not mapped (1)
[localhost:388268] Failing at address: 0x7ffe00000069
[localhost:388269] *** Process received signal ***
[localhost:388269] Signal: Segmentation fault (11)
[localhost:388269] Signal code: Address not mapped (1)
[localhost:388269] Failing at address: 0x7ffe00000069
[localhost:388268] [ 0] [localhost:388269] [ 0] /lib64/libc.so.6(+0x3cb20)[0x7fe57385fb20]
[localhost:388268] [ 1] /lib64/libc.so.6(+0x3cb20)[0x7f80f745fb20]
[localhost:388269] [ 1] /home/devel/mpi/openmpi/dev/lib/libopen-pal.so.0(ompi_sync_wait_mt+0x230)[0x7fe564d2da4e]
[localhost:388268] [ 2] /home/devel/mpi/openmpi/dev/lib/libopen-pal.so.0(ompi_sync_wait_mt+0x230)[0x7f80e8ac2a4e]
[localhost:388269] [ 2] /home/devel/mpi/openmpi/dev/lib/libmpi.so.0(+0x99c5d)[0x7fe564899c5d]
[localhost:388268] [ 3] /home/devel/mpi/openmpi/dev/lib/libmpi.so.0(+0x99c5d)[0x7f80e8299c5d]
[localhost:388269] [ 3] /home/devel/mpi/openmpi/dev/lib/libmpi.so.0(ompi_request_default_wait+0x27)[0x7f80e8299e8d]
[localhost:388269] [ 4] /home/devel/mpi/openmpi/dev/lib/libmpi.so.0(ompi_request_default_wait+0x27)[0x7fe564899e8d]
[localhost:388268] [ 4] /home/devel/mpi/openmpi/dev/lib/libmpi.so.0(MPI_Wait+0x138)[0x7f80e833f066]
...
[localhost:388269] *** End of error message ***
/lib64/libpython3.11.so.1.0(+0x1fce95)[0x7fe573bfce95]
[localhost:388268] *** End of error message ***
--------------------------------------------------------------------------
prterun noticed that process rank 0 with PID 0 on node localhost exited on signal 11 (Segmentation fault).
--------------------------------------------------------------------------
```

bosilca:
@dalcinl there was a small fix pushed related to ishrink, hopefully it fixes this issue for you because I am unable to replicate. Btw, I don't see any ishrink tests in this PR? Do you have some extra tests you are running or I am missing a commit ?

dalcinl:
@bosilca Yes, you were missing a commit. Now I pushed to your branch. 
I manually triggered a [build](https://github.com/mpi4py/mpi4py-testing/actions/runs/4297859800) of `bosilca/mpi4py@topic/ulfm` against current `open-mpi/ompi@main`. Let see how it goes.

dalcinl:
@bosilca Looks like the issue is still [there](https://github.com/mpi4py/mpi4py-testing/actions/runs/4297859800/jobs/7491342287).

PS: This happens with np=2, singleton or np=1 is fine. Also, take into account that I'm not passing `--with-ft` to mpiexec to run these tests. However, no failure is expected, so `Ishrink()` should return an output communicator congruent with the input communicator. Also note that mpi4py sets the error handler of the output communicator BEFORE waiting in the request (as it is done for `Idup()` ). Not sure if this could be a problem, but just in case I let you know.

abouteiller:
The C-based test https://github.com/ICLDisco/ulfm-testing/blob/ishrink/api/ishrink.c passes with faults.  I need to double check that the case of --with-ft=no  is handled correctly.

abouteiller:
I cannot replicate from C testers.

dalcinl:
@bosilca Sorry, I updated your branch with rebase using the GitHub UI. That process destroyed your commit signatures. 
I'm not sure what to do regarding this, I usually merge PRs using rebase, my guess is that GitHub will not preserve signatures in the final merge. Another option would be giving you temporary (or permanent) write access to the repo, then you can click the merge button yourself.

PS: I'll use this mess up as an opportunity to fixup the last commit in the previous two, so please don't re-signoff your commits just yet.

bosilca:
I can resign my commit once you're done. Meanwhile let me check what thread multiple changes for the handling of the ishrink request.

dalcinl:
@abouteiller @bosilca Any news? Could you take a look into the issue?

dalcinl:
Still failing with `ompi/main` [[logs]](https://github.com/mpi4py/mpi4py-testing/actions/runs/4637699411/jobs/8206823688#step:19:3132).

dalcinl:
@abouteiller @bosilca I'm trying to add more test to get 100% code coverage. In the process, I realized that we have not added any tests for **intercommunicators**. Moreover, I'm getting additional failures (segfaults) with current Open MPI main branch with some of the ULFM routines on intercommunicators (agree/iagree and shrink/ishrink).

dalcinl:
@bosilca I would like to get this PR merged. Please re-sign your commits and I'll perform a manual merge and push in the command line to preserve your signatures.

PS: The ishrink issue remains, plus a bunch of issues related to using intercommunicators. For the time being, mpi4py just skips these tests as know failures.

dalcinl:
@bosilca Many thanks! I'll merge as soon all CI builds pass.

@hzhou Unfortunately, I had to disable ULFM with MPICH for now. There are various issues with it (segfaults, missing support for intercommunicators, etc.). I'll try to follow up in with an issue later.

PR #329, opened 2023-03-15
build: Intel MPI and Python DLL search path on Windows

dalcinl:
Since Python 3.8, Windows DLLs are no longer searched in the `PATH` environment variable. While I understand the rationale behind this change, they should have introduced a new environment variable, let say `PYTHONDLLPATH`, to easily add to the search path. Otherwise, the only way is to use `os.add_dll_directory()` programatically.

Intel MPI does not install DLLs in system locations but the `I_MPI_ROOT` install tree. Moreover, the package installs two DLLs, one for debug and another for release. Users typically pick the one they want by running `setvars.bat`, and that batch script simply modifies `PATH`. But Python no longer uses `PATH` for DLL search. So here we are.

A quick and dirty workaround is to install a *.pth file in site-packages that will first look for `I_MPI_ROOT` defined in the environment, and if found, then call `os.add_dll_directory()` adding the `I_MPI_ROOT\bin\{debug|release}` directory to the DLL search path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -236,15 +236,21 @@ jobs:
         mpi:
           - mpich
           - openmpi
-        py:
-          - 3
+          - intelmpi
+        exclude:
+          - os:  macos-12
+            mpi: intelmpi
+          - os:  macos-11
+            mpi: intelmpi
         include:
-           - os: windows-2022
-             mpi: msmpi
-             py: 3
-           - os: windows-2019
-             mpi: msmpi
-             py: 3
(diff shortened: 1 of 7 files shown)


dalcinl:
@leofang @minrk As Microsoft MPI is abandonware, at some point we need to consider using Intel MPI on Windows. The issue discussed in here may be solved in a different way in conda-forge if we ever manage to package Intel MPI. In any case, removing `conf/intelmpi.pth` from the mpi4py source tree is enough for a conda recipe to get rid of the workaround I'm adding.

PR #340, opened 2023-04-05
mpi4py.util.pool: New drop-in replacement for multiprocessing.pool

dalcinl:


docs/source/mpi4py.util.pool.rst
```diff
@@ -0,0 +1,75 @@
...
+   :synopsis: :mod:`multiprocessing.pool` interface via :mod:`mpi4py.futures`.
+
+.. versionadded:: 4.0.0
+
+.. seealso::
+
+   This module intends to be a drop-in replacement for the
+   :mod:`multiprocessing.pool` interface from the Python standard library.
+   The :class:`~mpi4py.util.pool.Pool` class exposed here is implemented as a
+   thin wrapper around :class:`~mpi4py.futures.MPIPoolExecutor`.
```
leofang, line 14: Do people need to care about the implementation detail?

docs/source/mpi4py.util.pool.rst
```diff
@@ -0,0 +1,75 @@
...
+
+   .. automethod:: apply_async
+
+   .. automethod:: map
+
+   .. automethod:: imap
+
+   .. automethod:: imap_unordered
+
+   .. automethod:: map_async
```
leofang, line 38: nit: follow the same order as in Python docs
```suggestion
   .. automethod:: map_async

   .. automethod:: map

   .. automethod:: imap_unordered
```

docs/source/mpi4py.util.pool.rst
```diff
@@ -0,0 +1,75 @@
...
+
+   .. automethod:: starmap
+
+   .. automethod:: istarmap
+
+   .. automethod:: istarmap_unordered
+
+   .. automethod:: starmap_async
```
leofang, line 46: same nit
```suggestion
   .. automethod:: starmap

   .. automethod:: starmap_async

   .. automethod:: istarmap

   .. automethod:: istarmap_unordered
```

docs/source/mpi4py.util.pool.rst
```diff
@@ -0,0 +1,75 @@
...
+
+.. autoclass:: mpi4py.util.pool.AsyncResult
+
+   .. automethod:: ready
+
+   .. automethod:: successful
+
+   .. automethod:: wait
+
+   .. automethod:: get
```
leofang, line 67: same nit
```suggestion
   .. automethod:: get

   .. automethod:: wait

   .. automethod:: ready

   .. automethod:: successful
```

src/mpi4py/util/pool.py
```diff
@@ -0,0 +1,328 @@
...
+            if num_tasks == 0:
+                iterable = list(iterable)
+                num_tasks = len(iterable)
+            if num_tasks > 0:
+                quot, rem = divmod(num_tasks, num_workers * 4)
+                chunksize = quot + bool(rem)
+    return iterable, chunksize
+
+
+def _async_future_callback(future, callback=None, error_callback=None):
```
leofang, line 289: I am sensing the standalone helpers `_async*()` are here because you don't want any private methods to appear in the `Pool` class (which I am fine with)?

leofang (approved):
Did another pass after this PR came up. LGTM.

docs/source/mpi4py.util.pool.rst
dalcinl: The order here in docs is the same as the order the methods are defined in the source code. 
Right now the order is `foo`, `ifoo[_unordered]`, `foo_async`.
You are proposing a swappint to `foo`, `foo_async`, `ifoo[_unordered]`.
I think it is ultimately a matter of taste, but perhaps your way is easier to follow / reason about.

docs/source/mpi4py.util.pool.rst
```diff
@@ -0,0 +1,75 @@
...
+
+
+.. autoclass:: mpi4py.util.pool.ApplyResult
+   :show-inheritance:
+
+
+.. autoclass:: mpi4py.util.pool.MapResult
+   :show-inheritance:
```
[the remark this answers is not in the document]
dalcinl: One time I write comprehensive docs and you ask me to get rid of things?? Now seriously, Python docs are not necessarily prime examples of quality. They document `AsyncResult` as a class, but it is actually an alias, because they use to have a documentation bug. I prefer to keep them documented. Remember mpi4py provide typing stubs. Smart IDEs may show that `apply_async()`/`map_async()` return `ApplyResult`/`MapResult`, so it is good to have an entry in the docs for them.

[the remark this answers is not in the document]
leofang: Good point on typing. I haven't thought about that.

src/mpi4py/util/pool.py
dalcinl: In the most part, yes, I have a tendency to avoid private methods. But IMO, the only one that would make sense to add as a private method is `_chunksize()`. All the others are related to nasty implementation details that I prefer to keep as hidden as possible.

docs/source/mpi4py.util.pool.rst
dalcinl: My point here was to somehow diminish the old multiprocessing API and direct readers to the more powerful and modern futures API. Perhaps I should have been more vehement about it, although I'm being in the note following this sentence. I believe a power user/developer - which are the ones I really care about - would appreciate a succinct and high level comment about implementation.

docs/source/mpi4py.util.pool.rst
dalcinl: I've reordered things as per your suggestion.

docs/source/mpi4py.util.pool.rst
leofang: Sounds good. Though you do also have a note right below advertising the `futures` API, which already serves your goal.

docs/source/mpi4py.util.pool.rst
dalcinl: I have not yet changed the order of these ones. But I think I'm going to so it, under the assumption that `get()`/`wait()` are the most used/useful methods.

dalcinl: Done.

PR #352, opened 2023-04-29
Add support for a callback function in MPIPoolExecutor

adrn:
I often have a use case where I would like to run a callback function on the main process with the results from each worker process in an MPI pool. For example, to control file I/O, so that only one process is writing to a file. I have a subclass that implements this for my own use, but I thought I would check to see if you would be interested in pulling this feature upstream -- no worries either way. Thanks!

### DIFF
--- a/src/mpi4py/futures/pool.py
+++ b/src/mpi4py/futures/pool.py
@@ -120,7 +120,7 @@ class MPIPoolExecutor(Executor):
         submit.__text_signature__ = '($self, fn, /, *args, **kwargs)'
 
     def map(self, fn, *iterables,
-            timeout=None, chunksize=1, unordered=False):
+            callback=None, timeout=None, chunksize=1, unordered=False):
         """Return an iterator equivalent to ``map(fn, *iterables)``.
 
         Args:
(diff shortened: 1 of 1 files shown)


dalcinl:
I've been quite hesitant to extend the mpi4py.futures API beyond the one of concurrent.futures, although I've broke that rule (e.g., the `unordered` mode for `map`/`starmap`).

I have two questions/observations:

1. What's wrong with simply doing things the following trivial way?
```python
for res in executor.map(fn, iterable):
    callback(res)
```

2. How does your subclass-based solution looks like? I believe the special feature can be implemented in a just a few lines.
```python
from mpi4py.futures import *

class MyExecutor(MPIPoolExecutor):

    def map(self, fn, *iterables, **kwargs):
        callback = kwargs.pop('callback', None)
        iterator = super().map(fn, *iterables, **kwargs)
        if callback is not None:
            return ((res, callback(res))[0] for res in iterator)
        else:
            return iterator

def square(x):
    return x * x

if __name__ == '__main__':
    with MyExecutor(2) as executor:
        for _ in executor.map(square, range(5), callback=print):
            pass
```

So far, I'm not convinced this feature is worth pulling upstream.

adrn:
Nice, ok your subclass solution is simpler than mine - thanks! And sounds good.

PR #355, opened 2023-05-03
feature: Change a few Group/Comm class methods to instance methods

dalcinl:


leofang:
LGTM though I am not sure I understand the benefit 😄

dalcinl:
You can argue it is just about style. `g1.Compare(g2)` is shorter than `MPI.Group.Compare(g1, g2)`. Now you can choose to code it either way, both will work. Before this PR, only the second could be used.

I'm trying to remove unneeded class methods. I'm thinking about implementing the same change for group union/intersection/difference. Look at the builtin `set` class, its union/intersection/difference methods are instance methods, not class methods.  Class methods should be used when strictly necessary, e.g., factory methods.

leofang:
Thanks, Lisandro. Yes, that's a wise design choice.

PR #369, opened 2023-06-15
mpi4py.futures: Support for parallel tasks

dalcinl:


dalcinl:
@jan-janssen It took a while to make it bullet-proof, but here you have. You should take a look and provide some feedback. I added some documentation: [[link]](https://mpi4py--369.org.readthedocs.build/en/369/mpi4py.futures.html#parallel-tasks) [[link]](https://mpi4py--369.org.readthedocs.build/en/369/mpi4py.futures.html#cpi-py). Does it look good enough?

jan-janssen:
@dalcinl I am sorry for the delay, I was on holidays last week, but the functionality looks very promising that is exactly what I was looking for. Btw. do you already have a release date in mind for the next `mpi4py` version?

dalcinl:
#371 is another piece to the puzzle, as it fixes a limitation for  legacy MPI-1 (non-spawn) mode. BTW, The initializers can also be parallel and follow the similar rules as parallel tasks. 

There is only one thing I would like to implement: coalescing many futures (like the ones that you get submitting parallel tasks) in a single one, such that `future.result()` return a list of results. it would work roughly as:
```python
future = coalesce(iterable_of_futures)
list_of_results = future.result()
```

Regarding next release date, I've been waiting for Open MPI to release v5.0.0 with partial support for MPI-4. The next mpi4py release will incur in some minor performance penalties for MPI-3 implementations. I would rather prefer to wait until Open MPI v5 gets out and is readily available in at least rolling packages managers like Homebrew, conda, spack, etc.

jan-janssen:
I guess I am still a bit confused about the `MPI-2` mode. I always thought you divide the communicators while always keeping the same MPI ranks. Now as far as I understand `MPI-2` mode it uses mpi spawn to create more MPI ranks, correct? I am wondering how this works with queuing systems like SLURM which have their own wrappers like `srun`. For example I could ask for two nodes and then start two MPI parallel programs each using one node. How do you communicate mpi spawn which MPI ranks to use for which program?

dalcinl:
> I always thought you divide the communicators while always keeping the same MPI ranks.

That's what `mpi4py.futures` do in legacy MPI-1 mode. `MPI_COMM_WORLD` is split in one master and many workers. You can create multiple executors, but all share the same worker pool.

> Now as far as I understand `MPI-2` mode it uses mpi spawn to create more MPI ranks, correct? I am wondering how this works with queuing systems like SLURM which have their own wrappers like `srun`. 

When using `mpi4py.futures` in "modern" MPI-2 mode (spawn-based), you are supposed allocate many nodes but to `srun` only one MPI process running Python. Upon the creation of a `MPICommExecutor`, new MPI processes will be spawned to form the worker pool. At the time I implemented `mpi4py.futures`, the MPI implementation in the supercomputing facility I have access to used SLURM but did not support `MPI_Comm_spawn()`. Therefore,  I have had no chance to test this mode myself in production.

jan-janssen:
I am still a bit confused, let's say I have 10 cores and 100 sets of parameters I want to execute, with each function call executing one set of parameters on one core. How many `MPI_Comm_spawn()` commands do you call? 10 or 100? One for each core or one for each function call?

dalcinl:
Under MPI-2 mode, it all depends on how many executors you create. For each executor you create the is a corresponding `MPI_Comm_spawn()` call. Each spawn call will create `max_workers` new workers in one shot.

I'm afraid we are having trouble understanding each other. I tend to think in a rather general way albeit under the design constraints, and the fact that mpi4py.futures is primarily a package for embarrassingly parallel applications (i.e., sequential tasks), while you are focused in the particularities of your application and objectives, which involve using parallel tasks.

Rather than keeping adding comments to this merged PR, I would advise to open a new discussion, with some specific example, perhaps based in the compute-pi example I recently added. That way the requirement can be stated beforehand, and I can provide a proper solution.

PR #375, opened 2023-07-12
Test MPI-4.0 API coverage

dalcinl:


dalcinl:
@leofang With the test I'm adding in this PR, now we can claim with certainty that mpi4py covers 100% of the MPI-4.0 API.

leofang:
cc: @jeffhammond for vis

PR #376, opened 2023-07-19
refactor: Update to Cython 3

dalcinl:


setup.py
```diff
@@ -187,13 +187,9 @@ def extensions():
...
     if sys.version_info[:2] > maxknow_python:
+        api = '0x{:02x}{:02x}0000'.format(*maxknow_python)
         MPI['define_macros'].extend([
-            ('CYTHON_FAST_PYCALL', 0),
-            ('CYTHON_FAST_THREAD_STATE', 0),
-            ('CYTHON_USE_DICT_VERSIONS', 0),
-            ('CYTHON_USE_PYLONG_INTERNALS', 0),
-            ('CYTHON_USE_PYLIST_INTERNALS', 0),
-            ('CYTHON_USE_UNICODE_INTERNALS', 0),
+            ('CYTHON_LIMITED_API', api),
```
leofang, line 192: ! 

Does this work reliably now?

dalcinl: This is not the same as `Py_LIMITED_API`, it basically means _Don't use private CPython APIs_. 
The Python 3.12 failures in CI are related to enabling this flag. I found the root of the problem, have to submit an issue/PR to Cython.

conf/cyautodoc.py
leofang, line 1: This file is gone because it's upstreamd?

dalcinl: yes

leofang: Awesome, seems to be https://github.com/cython/cython/pull/5415.

setup.py
dalcinl: So I disabled `CYTHON_LIMITED_API` for Python 3.12, now our CI passes. 
I submitted a  PR to fix Cython: cython/cython#5549.

PR #379, opened 2023-07-20
type: Do not install stubs for Python 3.7 or older

dalcinl:
Python 3.7 is too old (and EOLed) and proper testing of typing stubs in CI would require workarounds to silence errors.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -49,7 +49,7 @@ Release 4.0.0 [2023-XX-XX]
 * Backward-incompatible changes:
 
   * Python 2 is no longer supported, Python 3.6+ is required, but
-    typing stubs are supported for Python 3.7+.
+    typing stubs are supported for Python 3.8+.
 
   * The `Intracomm.Create_group()` method is no longer defined in the
     base `Comm` class.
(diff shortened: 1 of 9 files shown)


leofang:
How about just dropping PY37? 🙂

dalcinl:
> How about just dropping PY37?

* Ubuntu 18.04 has Python 3.6 as default.
* Debian 10 has Python 3.7 as default.

There are machines running Ubuntu 18.04 in my workplace. 
I have a Raspberry Pi running Raspbian 10 in my own home.

leofang:
Yeah I know, but a common counterargument (which I am sure you've heard of) is that a user can use `pyenv` or `conda` to get a different Python version so as to not bound by system Pythons. What's your take? (Just curious!)

dalcinl:
In general lines, I'm just trying to avoid dependency hell. 

1. Thanks to Cython, supporting slightly dated Python releases is not a huge deal.
2. Asking users to use `pyenv` or `conda` is fine, but that assumes that users understand how to use these tools properly. You have to remember we are geeks, and not everyone using computers have the level of literacy required to fiddle with shells and virtual environment installers, and many times messy setup of HPC facilities. Yet, those computer-illiterate folks are doing science, and they need parallel processing capabilities.
3. Being able to use your system Python installation without having to rush to download some other thing is very convenient. I've seen first hand cases where users are stuck with a dated Python installation bundled within a larger packaged application, and they need to install additional packages to do post processing.

Yes, start using the walrus operator from 3.8 would be nice. But I'm not desperate for it, and I can very well wait a couple of Python release cycles to bump mpi4py's `python_requires`.

PR #402, opened 2023-09-18
feature: Add handle attribute and fromhandle() class method

dalcinl:


rainwoodman:
Nice! When will this get into a release? It will help us to fix https://github.com/rainwoodman/pmesh/issues/48 (Since we will be able to bind back the 2D communicator to Python)

dalcinl:
> When will this get into a release? 

Well, it will be available in the next 4.0.0 release, but for that to happen, you'll have to wait at least until Open MPI releases v5.0.0. 

> It will help us to fix rainwoodman/pmesh#48 (Since we will be able to bind back the 2D communicator to Python)

You do not have to wait until next release. 
In the mean time, you could just use the following `fromhandle` utility routine:

```python
from mpi4py import MPI

def fromhandle(mpi_class, handle):
    import ctypes as ct
    if MPI._sizeof(mpi_class) == ct.sizeof(ct.c_int):
        mpi_handle_p = ct.POINTER(ct.c_int)
    else:
        mpi_handle_p = ct.POINTER(ct.c_void_p)
    mpi_obj = mpi_class()
    address = MPI._addressof(mpi_obj)
    pointer = ct.cast(address, mpi_handle_p)
    pointer[0] = handle
    return mpi_obj


def _test(obj):
    handle = MPI._handleof(obj)
    dupe = fromhandle(type(obj), handle)
    assert obj == dupe
    assert obj is not dupe

    
if __name__ == '__main__':
    _test(MPI.COMM_NULL)
    _test(MPI.COMM_SELF)
    _test(MPI.COMM_WORLD)
    _test(MPI.DATATYPE_NULL)
    _test(MPI.INT)
    _test(MPI.DOUBLE)
    _test(MPI.FILE_NULL)
```

Issue #386, opened 2023-08-23
Next release? 4.0.0?


jakirkham:
It looks quite a bit of work has been done since the last release 👏

Am curious if there's a rough idea as to when the next release might come out. No pressure though (can see there's a lot of additions being included)

dalcinl:
I've been waiting for weeks and months for Open MPI to release 5.0.0 with (partial) support for MPI 4.0.

jakirkham:
Ah ok. That makes sense. Thanks for the context Lisandro! 🙏

Are there additional changes planned for mpi4py as part of this release or is it just\* a matter of upstream releases and testing?

\* - Recognizing there still may be considerable work to test and align

dalcinl:
1. I have a branch with some minor utility additions to `mpi4py.futures` to help with the new support for parallel MPI tasks. This branch has not been merged yet because I not sure about the naming of the two utility functions I'm adding and, as usual, documentation has to be improved.
2. I'm working on producing binary wheels for all platforms and architectures, ready for `pip install`. This is a daunting task. I cannot just follow the simple approach of other wheel distributors, that is, shipping shared libraries within the wheel. I rather have to skip including MPI libraries, rely on users setting LD_LIBRARY_PATH, and support the combination of MPI 3.1/4.0 standard for MPICH/OpenMPI ABIs. I'm about to finish, but it took way more time than anticipated. Binaries are here: https://anaconda.org/mpi4py/repo. I still have to open a new Issue/Discussion to advertise this work and get feedback. I still have to test whether the binaries actually work in Cray (via `module load cray-mpich-abi`).

jakirkham:
Thanks for the additional context! 🙏

balay:
@dalcinl  perhaps consider `3.2` release - for `cython-3` changes for now?

One reason for this is to have a compatible mpi4py with `petsc4py@3.20` - thats scheduled for sept end.

dalcinl:
@balay I'm considering making a new mpi4py release 3.1.5 to support Python 3.12. In the process, I will bound the Cython dependency below 3.0.0. The tarball will ship with fresh Cython-generated C sources, as it always did. But I'll not be updating the Cython *.pyx sources for Cython 3. I don't think my plan will affect spack, although I may be messing something.

balay:
>  I don't think my plan will affect spack, although I may be messing something.

@dalcinl As per discussion at https://github.com/spack/spack/pull/36406#discussion_r1147956576
[...]

PR #410, opened 2023-10-12
Minor improvements install instructions and documentation

### LINKED ISSUE #407: error: Cannot compile MPI programs. Check your configuration!!!
Hi there,

when installing mpi4py from PyPI (`pip install mpi4py`) users get the error `error: Cannot compile MPI programs. Check your configuration!!!` if they do not have a working MPI implementaion installed. There is no any warning or note both on README and on mpi4py's PyPI page as to the required prerequisites. I was wondering if we could add a note/warning regarging installation of mpi4py from PyPI, which says something like `you have to have a working MPI implementation before installing mpi4py`.

dalcinl:
Fixes #407.

cc: @YarShev @leofang

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -25,8 +25,8 @@ Interface* (`MPI <https://www.mpi-forum.org/>`_) standard. It is
 implemented on top of the MPI specification and exposes an API which
 grounds on the standard MPI-2 C++ bindings.
 
-Dependencies
-------------
+Prerequisites
+-------------
 
 * `Python <https://www.python.org/>`_ 3.6 or above,
   or `PyPy <https://www.pypy.org/>`_ 7.2 or above.
--- a/conf/mpidistutils.py
+++ b/conf/mpidistutils.py
@@ -908,13 +908,41 @@ def configure_mpi(ext, config_cmd):
       return 0;
     }
     """)
-    errmsg = "Cannot {} MPI programs. Check your configuration!!!"
+    errmsg = [
+        "Cannot {} MPI programs. Check your configuration!!!",
+        "Installing mpi4py requires a working MPI implementation.",
+    ]
+    if sys.platform == 'linux':
+        errmsg += [
+            "If you are running on a supercomputer or cluster, check with",
+            "the system administrator or refer to the system user guide.",
+            "Otherwise, if you are running on a laptop or desktop computer,",
+            "your may be missing the MPICH or Open MPI development package:",
+            "* On Fedora/RHEL systems, run:",
+            "  $ sudo dnf install mpich-devel     # for MPICH",
+            "  $ sudo dnf install openmpi-devel   # for Open MPI",
+            "* On Debian/Ubuntu systems, run:",
+            "  $ sudo apt install libmpich-dev    # for MPICH",
+            "  $ sudo apt install libopenmpi-dev  # for Open MPI",
+        ]
+    if sys.platform == 'darwin':
+        errmsg += [
+            "Install MPICH or Open MPI with Homebrew or MacPorts:"
+            "  $ brew install mpich|openmpi  # Homebrew",
+            "  $ port install mpich|openmpi  # MacPorts",
+        ]
+    if sys.platform == 'win32':
+        errmsg += [
+            "Please install *Intel MPI* or *Microsoft MPI*."
+        ]
     ok = config_cmd.try_compile(ConfigTest, headers=headers)
     if not ok:
-        raise DistutilsPlatformError(errmsg.format("compile"))
+        message = "\n".join(errmsg).format("compile")
+        raise DistutilsPlatformError(message)
     ok = config_cmd.try_link(ConfigTest, headers=headers)
     if not ok:
-        raise DistutilsPlatformError(errmsg.format("link"))
+        message = errmsg[0].format("link")
+        raise DistutilsPlatformError(message)
     #
     log.info("checking for missing MPI functions/symbols ...")
     impls = ("OPEN_MPI", "MSMPI_VER")
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -1,7 +1,8 @@
 Using **pip**
 -------------
 
-You can install mpi4py from its source distribution using ``pip``::
+You can install the latest mpi4py release from its source distribution
+at `PyPI <https://pypi.org/project/mpi4py/>`_ using ``pip``::
 
   $ python -m pip install mpi4py
 
@@ -15,8 +16,10 @@ or::
 
 .. note::
 
-   Installing mpi4py from sources requires a C compiler and an MPI
-   implementation with development headers and libraries.
+   Installing mpi4py from its source distribution (available at PyPI)
+   or Git source code repository (available at GitHub) requires a C
+   compiler and a working MPI implementation with development headers
+   and libraries.
 
 .. warning::
 


conf/mpidistutils.py
```diff
@@ -908,13 +908,38 @@ def configure_mpi(ext, config_cmd):
...
+    if sys.platform == 'linux':
+        errmsg += [
+            "You may be missing the MPICH or Open MPI devel package:",
+            "* On Fedora/RHEL:",
+            "  $ sudo dnf install mpich-devel     # for MPICH",
+            "  $ sudo dnf install openmpi-devel   # for Open MPI",
+            "* On Debian/Ubuntu:",
+            "  $ sudo apt install libmpich-dev    # for MPICH",
+            "  $ sudo apt install libopenmpi-dev  # for Open MPI",
+        ]
```
leofang, line 924:
```suggestion
            "  $ sudo apt install libopenmpi-dev  # for Open MPI",
            "* Or check with your system administrator / refer to the user guide if using a cluster."
        ]
```

conf/mpidistutils.py
```diff
@@ -908,13 +908,38 @@ def configure_mpi(ext, config_cmd):
...
+    if sys.platform == 'darwin':
+        errmsg += [
+            "Install MPICH or Open MPI with Homebrew or MacPorts:"
+            "  $ brew install mpich|openmpi  # Homebrew",
+            "  $ port install mpich|openmpi  # MacPorts",
+        ]
```
leofang, line 930: Q: Why is this needed for macOS? Didn't both of them provide binary packages?
- https://formulae.brew.sh/formula/mpi4py
- https://ports.macports.org/port/py-mpi4py/details/

So same as with conda, we shouldn't need special instruction?

README.rst
```diff
@@ -25,8 +25,8 @@ Interface* (`MPI <https://www.mpi-forum.org/>`_) standard. It is
 implemented on top of the MPI specification and exposes an API which
 grounds on the standard MPI-2 C++ bindings.
 
-Dependencies
-------------
+Prerequisites
+-------------
```
leofang, line 29: I have to say this is really like babysitting, I see no value added, but now that you've made the change I wouldn't even spend a second to argue 🙂

leofang (approved):

conf/mpidistutils.py
dalcinl: And Leo Fang just proved my point... whatever you write, is never enough, there is always something extra to say...

conf/mpidistutils.py
leofang: Fair enough, let me add more code suggestions.

README.rst
dalcinl: The new word sounds more pretentious... isn't that enough of a good reason for the change?

conf/mpidistutils.py
leofang:
```suggestion
    if sys.platform == 'darwin':
        errmsg += [
            "Install MPICH or Open MPI with Homebrew, MacPorts, or Conda:"
            "  $ brew install mpich|openmpi                  # Homebrew",
            "  $ port install mpich|openmpi                  # MacPorts",
            "  $ conda install -c conda-forge mpich|openmpi  # Conda",
        ]
```

conf/mpidistutils.py
leofang: (Suggestion updated.)

dalcinl: @leofang But your suggestion is wrong, you are missing the `{mpich|openmpi}-mpicc` package to pull in the compiler package, otherwise things will not work. Let me try to address it.

leofang: IIRC didn't we have bug reports for the mpicc wrappers from conda-forge?

conf/mpidistutils.py
```diff
@@ -908,13 +908,44 @@ def configure_mpi(ext, config_cmd):
...
+    if sys.platform == 'win32':
+        errmsg += [
+            "Please install *Intel MPI* or *Microsoft MPI*."
+        ]
```
leofang, line 940: I think msmpi is also available from conda-forge, not sure about intel mpi, though

conf/mpidistutils.py
dalcinl: Maybe, but I don't recall the context exactly. The fact is, the `mpicc` compiler wrapper and friends have the compiler command hardwired to the one used to build the MPI package. The <mpi>-mpicc packages are just convenience matapackages to install MPI + compiler toolchain. Without that, users get a confusing "command not found" error message.

conf/mpidistutils.py
dalcinl: If we keep this way we could very well dump the full install instructions. This is why I don't link writing documentation: no matter how much effort you put into it, someone will always have something to say about it, it is never enough.

conf/mpidistutils.py
dalcinl: Yes, not Intel MPI is also available. But think we are making a mistake mentioning conda-forge for the `pip install case`. An experienced user know what he/she is doing, a novice/beginner should probably stay away from mixing pip packages with conda packages.

conf/mpidistutils.py
leofang: Exactly. I am hoping to use this PR to justify to @yarshev that this effort really makes no sense. Read the documentation, we can't be held responsible for MPI vendors.

conf/mpidistutils.py
leofang: IIRC the problem was that conda-forge compiler does not play well, and if we bring in compiler toolchain from conda, instead of using the system one and just manually setting up `-I` and `-L`, we could run into issues.

dalcinl:
@leofang  On  a second though, I believe is better to remove references to conda and conda-forge. Let's assume that `pip` users are not installing mpi4py in a conda environment, and if they do, they know about `conda install mpi4py`, and they choose to use pip for a good reason, and therefore they know what they are doing. Isn't that reasonable?

conf/mpidistutils.py
leofang: Wouldn't this be the same case for brew/port?

leofang:
I think the whole PR should just be nuked and replaced by a big banner:
```
mpi4py does not distribute pre-compiled binaries (wheels) on PyPI, when you do
"""
pip install mpi4py
"""
it would download the mpi4py sdist (source distribution) and build from source.
For pre-compiled binaries, please check out the documentation of your favorite package manager (HomeBrew/MacPorts/Conda/apt/yum/...). 
```
I think it's what confused @yarshev.

conf/mpidistutils.py
dalcinl: Well, yes, you have a point. Now that I've spent so much time in it, let me do the full think at least for Linux and macOS. As usual, Windows may require more careful considerations I'm not willing to make for free.

conf/mpidistutils.py
dalcinl: So let's remove all references to conda then. I don't want a clarification that contributes to spread even more confusion.

YarShev (commented):
I think this looks much clearer now. I tried to install mpi4py from your branch.

```bash
pip install -e .
```

Got this error.

```bash
      error: Cannot compile MPI programs. Check your configuration!!!
      Installing mpi4py requires a working MPI implementation.
      If you are running on a supercomputer or cluster, check with
      the system administrator or refer to the system user guide.
      Otherwise, if you are running on a laptop or desktop computer,
      your may be missing the MPICH or Open MPI development package:
      * On conda environments with conda-forge packages, run:
        $ conda install -c conda-forge mpich-mpicc   # for MPICH
        $ conda install -c conda-forge openmpi-mpicc # for Open MPI
      * On Fedora/RHEL systems, run:
        $ sudo dnf install mpich-devel     # for MPICH
        $ sudo dnf install openmpi-devel   # for Open MPI
      * On Debian/Ubuntu systems, run:
        $ sudo apt install libmpich-dev    # for MPICH
        $ sudo apt install libopenmpi-dev  # for Open MPI
```

Based on the output I installed `mpich-mpicc`.

```bash
mamba install -c conda-forge mpich-mpicc
```

After that I was able to install mpi4py and ran a simple test.

The changes on `INSTALL.rst` also look good to me.

Many thanks! I hope this will help users a lot in installing mpi4py and depending on it packages.

dalcinl:
> Based on the output I installed `mpich-mpicc`.

Well, I eventually removed that stuff. If you are using conda, you should just `conda install mpi4py`. 
Users wanting to mix conda packages and pip packages are expected to know what they are doing.

YarShev:
Even after removing all references to conda, I think the changes are very valuable and much clearer for users.

YarShev (approved):

YarShev:
@dalcinl, could you shed light on which release will contain the changes? Is there some cadence for mpi4py releases?

dalcinl:
> could you shed light on which release will contain the changes? Is there some cadence for mpi4py releases?

See #386.

YarShev:
@dalcinl, I was wondering if we should add some lines about installation on Windows if possible?

```
      * On Fedora/RHEL systems, run:
        $ sudo dnf install mpich-devel     # for MPICH
        $ sudo dnf install openmpi-devel   # for Open MPI
      * On Debian/Ubuntu systems, run:
        $ sudo apt install libmpich-dev    # for MPICH
        $ sudo apt install libopenmpi-dev  # for Open MPI
```

dalcinl:
@YarShev I mean, why the `"Please install *Intel MPI* or *Microsoft MPI*." message is not enough?

YarShev:
@dalcinl, I wonder if there is a way (one-line command) to install an MPI implementation on Windows using some kind of a package manager for Windows as for Linux? We could put that line in the error message above.

dalcinl:
* There are many ways to install packages on Windows, from native installers you download from the MPI provider webpages directly, to other pkg managers like winget, nuget, chocolately, just to name the ones I've heard about. I'm not a Windows user, and I don not have enough expertise as to suggest others what they should use. Moreover, given that there are so many options, I'm not willing to dump a complete MPI installation manual just for the occasional clueless user. 
* Installing MPI is just one piece of the puzzle, users should install a MSVC, wich is a large download, and I'm not even sure it is available without subscribing to some MS developer account. Without a C compiler, building mpi4py from sources would fail miserably. Am I supposed to teach users how to deal with that mess, too? Where do we stop?
* For a some time, mpi4py has provided pre-built binary wheels built against Microsoft MPI. That may change for next release, as Intel MPI is also available in Windows, and provides support for the MPI 3.1 standard. Long story short, this means that `pip install mpi4py` on Windows will install a pre-built wheel, and there is no need to build from sources. Of course, installing Microsoft MPI (or Intel MPI in the future) is still a requirement.
* Finally, if you are still worried that your Windows consumers will struggle with MPI, maybe we should point them to the stuff [here](https://github.com/mpi4py/mpi4py-publish) and provide the one-line instruction about how to pip install Intel MPI and mpi4py (that is, the "(i) Note" box).

YarShev:
@dalcinl, I understand. Thank you for the comprehensive explanation. It seems to me that is a good point to add a link to the stuffe you mentioned in the error message - `"Please install Intel MPI or Microsoft MPI.". Of course, if it will not bloat up the error message.

leofang:
> Am I supposed to teach users how to deal with that mess, too? Where do we stop?

[My position on this](https://github.com/mpi4py/mpi4py/pull/410#issuecomment-1760242439) has been that
> I think the whole PR should just be nuked and replaced by a big banner

dalcinl:
This is a project of worldwide reach for a niche need with a very small dev team and zero funding. Every second I spend on new developments and maintenance is time I could use for tangible benefit in other activities. Yet, I do this work for personal fulfillment, to keep myself up-to-date and learn new thing, and to interact with professionals that know what they are doing and producing stuff many others benefit, either directly or indirectly. In this scheme of things, you like it or not, newbies have very low priority.

YarShev:
Makes sense, we can leave the error message as is.

PR #425, opened 2023-11-13
refactor: mpi4py.get_config() now returns an empty dictionary

dalcinl:
From next release, `mpi4py.get_config()` will return an empty dictionary. However, third-party packagers and distributors can still add an `mpi.cfg` file as package data to provide users with MPI  build information (compiler wrappers, includes, libraries).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,5 +17,4 @@
 /src/mpi4py/MPI.h
 /src/mpi4py/MPI_api.h
 /src/mpi4py/MPI.*.so
-/src/mpi4py/mpi.cfg
 /src/mpi4py.egg-info/**
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -85,6 +85,8 @@ Release 4.0.0 [2023-XX-XX]
 
   * The `mpi4py.dl` module is no longer available.
 
+  * The `mpi4py.get_config` function returns an empty dictionary.
+
 
 Release 3.1.5 [2023-10-04]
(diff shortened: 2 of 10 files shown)


dalcinl:
@leofang I remember long ago you asked about `mpi.cfg` and editable installs. This change may affect your workflow.

leofang:
> @leofang I remember long ago you asked about `mpi.cfg` and editable installs. This change may affect your workflow.

So I assume you meant #302, which was fixed by #304? Am I right that what part of this PR is doing is to undo that PR?

I have no problem. As long as get_config() does not throw me a KeyError under any circumstance (that was the original problem). You may wanna write a short snippet explaining to package managers how to create a valid mpi.cfg, though. But thanks for double checking 🙂

dalcinl:
> So I assume you meant #302, which was fixed by #304? 

Yes.

> Am I right that what part of this PR is doing is to undo that PR?

It's more than that. The `mpi.cfg` file is no longer generated, even with regular installs. 

> I have no problem. As long as calling get_config() does not throw me a KeyError under any circumstance (that was the original problem). 

Don't worry. As I explained in the deprecation notice, now the function returns an empty dictionary by default.

> You may wanna write a short snippet explaining to package managers how to create a valid mpi.cfg, though.

OK, I'll mention that `mpi.cfg` is in INI format to be read by the `configparser` module, and the MPI information has to be under the `[mpi]` section.

PR #433, opened 2023-11-22
bugfix: Rework is_integral check to handle NumPy arrays

### LINKED ISSUE #430: scount/rcount no longer work as np.arrays
Hi! 

I noticed that Scatterv and possibly other functions no longer accept np.arrays as their scount/rcount arguments in master branch.

<details><summary>Example</summary>
<p>

```python
#!/usr/bin/env mpiexec -n 3 python

import numpy as np
from mpi4py import MPI

comm = MPI.COMM_WORLD
assert comm.size == 3

amounts = np.array([3, 4, 2])  # | works as lists
offsets = np.array([0, 3, 7])  # |

if comm.rank == 0:
    data = np.arange(sum(amounts), dtype=float)
else:
    data = None

buffer = np.empty(amounts[comm.rank], dtype=float)

comm.Scatterv(
    [data, amounts, offsets, MPI.DOUBLE],
    [buffer, amounts[comm.rank], MPI.DOUBLE],
    root=0
)

print(f"{comm.rank}, {buffer =}")
```

</p>
</details>

It broke from #356, when function is_integral in src/mpi4py/MPI/msgbuffer.pxi was replaced with PyIndex_Check.
Is it intentional, or can we bring it back?

dalcinl:
Fixes #430

### DIFF
--- a/src/mpi4py/MPI/asarray.pxi
+++ b/src/mpi4py/MPI/asarray.pxi
@@ -1,5 +1,23 @@
 # -----------------------------------------------------------------------------
 
+cdef extern from "Python.h":
+    int PyIndex_Check(object)
+    int PySequence_Check(object)
+    object PyNumber_Index(object)
+    Py_ssize_t PySequence_Size(object) except -1
+
+cdef inline int is_integral(object ob) noexcept:
+    if not PyIndex_Check(ob):    return 0
+    if not PySequence_Check(ob): return 1
+    try: PySequence_Size(ob)
+    except: pass
+    else: return 0
+    try: PyNumber_Index(ob)
+    except: return 0
+    else: return 1
+
+# -----------------------------------------------------------------------------
+
 cdef extern from * nogil:
     const int INT_MAX
 
@@ -25,7 +43,8 @@ cdef inline object getarray(object ob, count_t *n, integral_t **p):
     if count_t is int: chklength(size)
     cdef integral_t *base = NULL
     cdef object mem = newarray(size, &base)
-    for i in range(size): base[i] = ob[i]
+    for i in range(size):
+        base[i] = PyNumber_Index(ob[i])
     n[0] = <count_t> size
     p[0] = base
     return mem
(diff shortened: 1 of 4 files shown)


dalcinl:
@s-valent In here, I'm implementing a slightly simpler version of the original `is_integral()` predicate. Hopefully, this will still play nice with NumPy arrays. Can you imagine of any corner case whether this would not work as expected?

dalcinl:
BTW, I believe the original check was so convoluted simply because of Pyhon 2 oddities. Now that mpi4py is Python 3 only, what I'm doing in this PR may be enough, although I'm not 100% sure.

s-valent:
It seems to be working differently for the case of 0-dimentional arrays (for example, `np.array(2)`), although I'm not sure if it's an issue or not generally

dalcinl:
What do you mean exactly? I'm not following. Do you have any examples where things do not work as expected?

s-valent:
> Do you have any examples where things do not work as expected?

I can't really say it works unexpectedly, just that the behavior changed

```
#!/usr/bin/env mpiexec -n 1 python

import numpy as np
from mpi4py import MPI

comm = MPI.COMM_WORLD

amount = np.array(1)
offset = np.array(0)

data = np.arange(amount, dtype=float)
buffer = np.empty(amount, dtype=float)
comm.Scatterv([data, amount, offset, MPI.DOUBLE], [buffer, amount, MPI.DOUBLE])
```

It's probably fine though, since rcounts and displs as 0-dim arrays are useless

dalcinl:
@s-valent Sorry for the noise. I'm not 100% focused, I missed the very important fact that CI was failing 🤣 . Looks like the only way to go is to reinstate the original implementation of the check. Let see how it goes. If CI passes and you have no further objections/comments, then I'll merge.

s-valent:
No worries, I can never bring myself to be focused)

I have no objections

dalcinl:
@s-valent I added some additional measures to prevent floating dtypes to convert silently as integral (that's how Cython handles things, following C semantics for the cast operator). This may trigger regressions in user codes. However, using floating point values as counts/offsets/indices is something we should not let pass silently and error. Moreover, such automatic conversions are no longer standard behavior in recent Python.

PS: Many thanks for your bug report. This would have been an embarrassing regression for next mpi4py release. No matter how much tests you add, there is always some corner case you are not triggering.

PR #442, opened 2023-12-14
Update after MPICH 4.3 CH4:UCX support for dynamic processes

dalcinl:
https://github.com/pmodels/mpich/pull/5467
https://github.com/pmodels/mpich/pull/6835
https://github.com/pmodels/mpich/pull/6822
https://github.com/pmodels/mpich/pull/6816

### DIFF
--- a/demo/futures/test_futures.py
+++ b/demo/futures/test_futures.py
@@ -1735,16 +1735,18 @@ if name == 'MPICH':
     if sys.platform == 'darwin':
         if version >= (3, 4) and version < (4, 0):
             SKIP_POOL_TEST = True
-    if MPI.COMM_WORLD.Get_attr(MPI.APPNUM) is None:
-        SKIP_POOL_TEST = (version < (4, 1))
-    try:
-        port = MPI.Open_port()
-        MPI.Close_port(port)
-    except:
-        port = ""
-    if port == "":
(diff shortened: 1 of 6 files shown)


demo/futures/test_futures.py
```diff
@@ -1735,16 +1735,18 @@ def test_except_hook_failure(self):
...
+            SKIP_POOL_TEST = True
+    if version < (4, 3):
+        try:
+            port = MPI.Open_port()
+            MPI.Close_port(port)
+        except:
+            port = ""
+        if port == "":
+            SKIP_POOL_TEST = True
+        del port
```
leofang, line 1749: Why not just do
```python
except:
    SKIP_POOL_TEST = True
```
?

dalcinl: Because in previous version of MPICH with `device=ch4:ucx` the port returned by Open_port() would actually be the (unusable) empty string, or alternatively the call would fail with exception, depending on whether you were using `pm=hydra` or `pm=gforker`. In any case, all these checks will become legacy. At this point, ch4:ucx supports everything mpi4py tests for. Some day we could switch to it in conda-forge. We will have to wait until MPICH 4.3, though.

leofang: I see, life is so complicated!

PR #503, opened 2024-05-28
CAPI: Support limited API and ABI compatibility

dalcinl:


leofang:
Interesting, I thought Cython has not officially supported limited API (cython/cython#2542, cython/cython#6082)? Is this already working today?

dalcinl:
@leofang This is "limited API" for mpi4py :-), now I realize the term is a bit confusing... This is basically for mpi4py users exchanging MPI handles across Python and C/C++ boundaries. It is a limited API because under it you cannot access in C the internals of the MPI objects (as declared in MPI.pxd). Instead, you just use some `new` and `get` set of routines, as done in some examples involving C/C++.

PR #509, opened 2024-06-16
Use BigMPI if MPI 4.0 is not available

chillenb:
I'm not sure when MPI 4.0 will become widely available, so I thought it might be a good idea to use BigMPI to fill in the gap until that happens. Seems like MPICH is the only implementation with support at the moment...

### DIFF
--- a/src/lib-mpi/bigmpi.h
+++ b/src/lib-mpi/bigmpi.h
@@ -1,5 +1,32 @@
+/*
+This code is adapted from Jeff Hammond's BigMPI.
+https://github.com/jeffhammond/BigMPI
 
-//#include <mpi.h>
+The MIT License (MIT)
+
+Copyright (c) 2013, 2014 Argonne National Laboratory
+Copyright (c) 2014       Friedrich-Alexander-Universitaet Erlangen-Nuernberg
+Copyright (c) 2014       Intel Corporation
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
 #include <stdint.h>
 #include <limits.h>
 
--- a/src/lib-mpi/bigmpi.h
+++ b/src/lib-mpi/bigmpi.h
@@ -1,0 +1,493 @@
+
+//#include <mpi.h>
+#include <stdint.h>
+#include <limits.h>
+
+#define BIGMPI_COUNT_MAX (SIZE_MAX)
+#define BIGMPI_INT_MAX (INT_MAX)
+#if ( defined(__GNUC__) && (__GNUC__ >= 3) ) || defined(__IBMC__) || defined(__INTEL_COMPILER) || defined(__clang__)
+#  define bmpi_unlikely(x_) __builtin_expect(!!(x_),0)
+#  define bmpi_likely(x_)   __builtin_expect(!!(x_),1)
+#else
+#  define bmpi_unlikely(x_) (x_)
+#  define bmpi_likely(x_)   (x_)
+#endif
+
+static int BigMPI_Type_contiguous(MPI_Aint offset, MPI_Count count, MPI_Datatype oldtype, MPI_Datatype * newtype)
+{
+    /* The count has to fit into MPI_Aint for BigMPI to work. */
+    if ((uint64_t)count>(uint64_t)BIGMPI_COUNT_MAX) {
+        return MPI_ERR_COUNT;
+    }
+
+    MPI_Count c = count/BIGMPI_INT_MAX;
+    MPI_Count r = count%BIGMPI_INT_MAX;
+
+
+    if(c>=BIGMPI_INT_MAX || r>=BIGMPI_INT_MAX){
+        return MPI_ERR_COUNT;
+    }
+
+    MPI_Datatype chunks;
+    MPI_Type_vector(c, BIGMPI_INT_MAX, BIGMPI_INT_MAX, oldtype, &chunks);
+
+    MPI_Datatype remainder;
+    MPI_Type_contiguous(r, oldtype, &remainder);
+
+    MPI_Aint lb /* unused */, extent;
+    MPI_Type_get_extent(oldtype, &lb, &extent);
+
+    MPI_Aint remdisp          = (MPI_Aint)c*BIGMPI_INT_MAX*extent;
+    int blocklengths[2]       = {1,1};
+    MPI_Aint displacements[2] = {offset,offset+remdisp};
+    MPI_Datatype types[2]     = {chunks,remainder};
+    MPI_Type_create_struct(2, blocklengths, displacements, types, newtype);
+
+    MPI_Type_free(&chunks);
+    MPI_Type_free(&remainder);
+
+    return MPI_SUCCESS;
+}
+
+static int MPIX_Type_contiguous_x(MPI_Count count, MPI_Datatype oldtype, MPI_Datatype * newtype)
+{
+    return BigMPI_Type_contiguous(0, count, oldtype, newtype);
+}
+
+
+static int MPIX_Send_x(void *buf, MPI_Count count, MPI_Datatype datatype, int dest, int tag, MPI_Comm comm)
+{
+    int rc = MPI_SUCCESS;
+
+    if (bmpi_likely (count <= BIGMPI_INT_MAX )) {
+        rc = MPI_Send(buf, (int)count, datatype, dest, tag, comm);
+    } else {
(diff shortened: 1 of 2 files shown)


chillenb:
I didn't include any RMA or reductions, but I'd be happy to do it if there's interest.

dalcinl:
I'm afraid I cannot accept this PR in its current form.

* The `largecnt.h` header file is generated from running `conf/mpiapigen.py`. This issue should be relatively trivial to address, though.
* The BigMPI API added here cover some MPI calls but not all. For example, "vector" collectives are not covered. I understand it is not possible in general to support everything. However, there are a few internal convention in mpi4py that should allow for everything to work. In fact, these conventions are currently allowing the translation of all large count APIs to the old small count ones.
* Many of the BigMPI APIs as added here are unfortunately broken for the case of intercommunicators. For the case of collectives, and particularly on intercommunicators, sometimes input arguments should be ignored. That's a deal breaker for me. I'm not sure whether this is an oversight from @jeffhammod, or he simply did not care about all the cases. After all, Jeff's Big MPI was an exercise to prove the MPI Forum that the "just use datatypes" approach could not possibly work in general.

When I updated mpi4py for MPI-4, I did consider adding Jeff's BigMPI stuff in mpi4py. However, after realizing all the little issues yet to address, and being under self-pressure to finish the MPI-4 support, I finally decided to leave the BigMPI support out. Additionally, I don't even know how reliable the BigMPI approach would be in practice: after all, we need MPI implementations to handle counts such that no internal integer overflow occurs. IIRC, Open MPI is working right now on stremlining their internals to support large count as a prerequisite to actually adding the new large count APIs.  Adding injury to insult, mpi4py testsuite is not even prepared to stress-test the large count path. Proof of that is that tests in Azure happily passed. However, trust me when I say that many tests should have failed bad provided the message counts would have been larger than `INT_MAX`.

All that being said, let me finish with two points:

* I'm not opposed at all to adding BigMPI to mpi4py. I didn't do it myself out of a time investment vs. benefit from my  very particular point of view. However, if BigMPI is ever added, it has to be as comprehensible as possible, and even more importantly, it has to be correct. A simple copy&paste of Jeff's BigMPI with a minor tweaks here and there will not work. Again, I do not haveIf you are still interested in pursuing this work, let me know and we can discuss the details.
* I'm about to release mpi4py 4.0.0, most likely this week, I cannot keep delaying it. If BigMPI support is ever added, it can get in an upcoming patch release.

dalcinl:
> I didn't include any RMA or reductions, but I'd be happy to do it if there's interest.

RMA accumulate routines  cannot be implemented in BigMPI, as they can only work with predefined MPI datatypes.

chillenb:
Thanks for your lengthy feedback. I appreciate it! I wasn't expecting you to merge this---rather, I was hoping for you to tell me about what could go wrong with this approach.

Would it make more sense to implement a subset of BigMPI in Python, to be published as a separate package and used on top of mpi4py? It would be less efficient than Cython, but many users might end up writing similar code themselves, so it'd save them a bit of work. 

> RMA accumulate routines cannot be implemented in BigMPI, as they can only work with predefined MPI datatypes.

Would you mind clarifying this statement for me---do you mean that the [RMA accumulate routines in BigMPI](https://github.com/jeffhammond/BigMPI/blob/master/src/rma_x.c) are faulty or are otherwise not going to work with the conventions in mpi4py? 

> Open MPI is working right now on stremlining their internals to support large count as a prerequisite to actually adding the new large count APIs. 

I hope they finish it soon! If you try to send a really big derived datatype in OpenMPI right now, could you cause an integer overflow? Isn't that a serious bug? Their OpenSHMEM implementation takes `size_t` counts---hopefully that's free of integer overflows? :(

> However, if BigMPI is ever added, it has to be as comprehensible as possible, and even more importantly, it has to be correct.

The BigMPI paper mentions that nonblocking v-collectives can't be implemented. Also, MPI_IN_PLACE isn't supported for user-defined reductions. If BigMPI support is added to mpi4py, would those limitations be acceptable?

dalcinl:
> I was hoping for you to tell me about what could go wrong with this approach.

Well, technically speaking, nothing can really go wrong if things are done properly. My only concern is just the following: is it worth all the implementation (and potentially maintenance) work?

> Would it make more sense to implement a subset of BigMPI in Python, to be published as a separate package and used on top of mpi4py? It would be less efficient than Cython, but many users might end up writing similar code themselves, so it'd save them a bit of work.
> 

Maybe yes, but then you may need to repeat a lot of the internal mpi4py logics to implement both buffer-based and pickle-based communication.

> > RMA accumulate routines cannot be implemented in BigMPI, as they can only work with predefined MPI datatypes.
> 
> Would you mind clarifying this statement for me---do you mean that the [RMA accumulate routines in BigMPI](https://github.com/jeffhammond/BigMPI/blob/master/src/rma_x.c) are faulty or are otherwise not going to work with the conventions in mpi4py?
> 

Sorry, I got confused. From the MPI standard: _Each datatype argument must be a predefined datatype or a derived datatype where all basic components are of the same predefined datatype`. I forgot about the second part of the OR. BigMPI datatypes certainly fit on these constraints.

> > Open MPI is working right now on stremlining their internals to support large count as a prerequisite to actually adding the new large count APIs.
> 
> I hope they finish it soon! If you try to send a really big derived datatype in OpenMPI right now, could you cause an integer overflow? Isn't that a serious bug? 

I don't really know. But the fact that they are doing this preparation work before going with the MPI-4 implementation rings a bell. We should ask them about it. If current and older Open MPI versions would have trouble with internal integer overflow under BigMPI, then this wold reinforce my "is it worth?" concern.

> Their OpenSHMEM implementation takes `size_t` counts---hopefully that's free of integer overflows? :(

Well, IIRC their OpenSHMEM implementation is unrelated to the MPI one, but I may be wrong


> > However, if BigMPI is ever added, it has to be as comprehensible as possible, and even more importantly, it has to be correct.
> 
> The BigMPI paper mentions that nonblocking v-collectives can't be implemented. 

What about BLOCKING v-collectives? I do not remember...

> Also, MPI_IN_PLACE isn't supported for user-defined reductions. 

This looks like problem. I don't care about the actual support for it, but rather whether you can still detect MPI_IN_PLACE is being used in some processes without extra collective communication. Such detection may be needed to implement the proper dispatch of the small vs large count case.

> If BigMPI support is added to mpi4py, would those limitations be acceptable?

Limitations is not my main concern, correctness is. Also not slowing down the small-count case.

Perhaps we should look first on how to implement BigMPI point to point in full, and once that is done, in a second stage we look at collectives and decide what can go truly go in under the correctness+performance constraint? Would such approach work for you?

jeffhammond:
> > I didn't include any RMA or reductions, but I'd be happy to do it if there's interest.
> 
> RMA accumulate routines cannot be implemented in BigMPI, as they can only work with predefined MPI datatypes.

No, that's not true.  Accumulate happens element-wise on the basis datatype elements of a user-defined datatype.  ARMCI-MPI has been doing Accumulate with user-defined datatypes for more than a decade.

dalcinl:
> > RMA accumulate routines cannot be implemented in BigMPI, as they can only work with predefined MPI datatypes.
> 
> No, that's not true.

Yes, sorry, I confused datatypes with user-defined reductions.

jeffhammond:
> > However, if BigMPI is ever added, it has to be as comprehensible as possible, and even more importantly, it has to be correct.
> 
> The BigMPI paper mentions that nonblocking v-collectives can't be implemented. Also, MPI_IN_PLACE isn't supported for user-defined reductions. If BigMPI support is added to mpi4py, would those limitations be acceptable?

Non-blocking v-collectives are possible, just difficult.  I figured out how to do them when I implemented https://github.com/jeffhammond/mukautuva.  The problem is that the implementation requires intercepting request completion, which is not trivial.

`MPI_IN_PLACE` is probably not that hard, I just didn't try to do everything in BigMPI because the purpose was to show the MPI Forum we needed to add large-count support to the standard, not be a perfect implementation.

jeffhammond:
> Limitations is not my main concern, correctness is. Also not slowing down the small-count case.

It's one branch in BigMPI to filter the small-count case.
 
> Perhaps we should look first on how to implement BigMPI point to point in full, and once that is done, in a second stage we look at collectives and decide what can go truly go in under the correctness+performance constraint? Would such approach work for you?

Do p2p and blocking collectives, both of which are trivial.  I'm not sure there's even a problem with `MPI_IN_PLACE`, I just haven't thought about it much.

dalcinl:
In private communication, @jeffhammond warned me that BigMPI never cared about user-defined datatypes. Looking at the BigMPI datatype is constructed, I'm not sure things would work for arbitrary user-defined datatypes. If that's the case, that's another point to consider on whether we want this stuff in mpi4py.

jeffhammond:
> In private communication, @jeffhammond warned me that BigMPI never cared about user-defined datatypes. Looking at the BigMPI datatype is constructed, I'm not sure things would work for arbitrary user-defined datatypes. If that's the case, that's another point to consider on whether we want this stuff in mpi4py.

It might just work, I just haven't thought about it.  As I say in the BigMPI docs, if you know how to create your own datatypes, you don't need BigMPI, because all BigMPI does is create a user-defined datatype that holds more than `INT_MAX` elements.

dalcinl:
> > Limitations is not my main concern, correctness is. Also not slowing down the small-count case.
> 
> It's one branch in BigMPI to filter the small-count case.
> 

I'm not talking about the cost of a branch. Rather, I'm worried about having to add extra collective MPI calls in order to achieve correctness (for example, to support `MPI_IN_PLACE`). That would not be acceptable.

jeffhammond:
i don't understand why a collective would be required to implement `MPI_IN_PLACE`.  as i said, i don't even know that it's broken.  i just didn't test it.  at this point, i don't intend to put in the effort to port the MPICH test suite to the BigMPI API so that i can test `MPI_IN_PLACE` (in emulation mode, where i artificially switch on the BigMPI protocol).

if you have tests that fail, i will happily look into fixing bugs in BigMPI.

dalcinl:
> i don't understand why an extra collective would be required to implement MPI_IN_PLACE

That comment was just pure speculation from my side. What I was trying to say is that: if MPI_IN_PLACE is broken, and fixing it requires some global synchronization, then such fix is not acceptable for mpi4py. 

> if you have tests that fail, i will happily look into fixing bugs in BigMPI.

As you know, BigMPI is very hard to test simply because you have to go over the 2GB limit. Look what happened in this PR: all tests passed with Open MPI, but that's simply because the small-count branch is being taken in all tests. I'm pretty sure that if you take the changes of this PR and set `bigmpi_int_max` to a very small value (let say just 2), then eventually some tests will fail (although it may be just the ones related to intercommunicators that you do not care).

dalcinl:
Superseded by #628, which only handles the minimum feature set we are sure works with predefined and user-defined datatypes, that is, BigMPI's `MPI_Type_contiguous_x`. 

From the Python side, `MPI.Datatype.Create_contiguous(count)` will create the right datatype corresponding to small vs large counts.

PR #511, opened 2024-07-06
test: Disable tests failing with Intel MPI 2021.13

dalcinl:


dalcinl:
@ddurnov @rafael-lago FYI

* The Windows datatype and IO issues have not been solved in latest release 2021.13. From now on, I'm unconditionally skipping these tests until new notice, I simply do not have the time to chasing down bugs and keep dealing with test failures. 
* Additionally, there are new weird regressions in non-blocking RMA put/get, accumulate, and get-accumulate. These errors cannot be reproduced running the individual tests on their own, but they do lead to segfaults when the whole test suite is run. Again, I'm unconditionally skipping these tests from 2021.13 and up until new notice.

PR #536, opened 2024-08-22
Add --xfile option to provide an exclude pattern file

ykhnygin-intel:
In case there is a significant number of exclude patterns, it would be more convenient to supply them in a line-separated file. I have implemented this feature here.


ykhnygin-intel:
@dalcinl It works on my side, thank you!

PR #611, opened 2025-02-12
dlpack: Do not call PyBuffer_Release on the exporter object

dalcinl:
Fixes #609 

cc @snsunx-nv @seberg

Issue #609, opened 2025-02-10
Double free or corruption error for collective operations on CuPy arrays


snsunx-nv:
When CuPy is installed from source on its latest `main` branch, mpi4py fails on a simple collective test:

```python
import os

import cupy as cp
from mpi4py import MPI


comm = MPI.COMM_WORLD
rank = comm.Get_rank()
size = comm.Get_size()

num_devs = cp.cuda.runtime.getDeviceCount()
# this assumes Open MPI
dev_id = (int(os.environ["OMPI_COMM_WORLD_LOCAL_RANK"]) % num_devs)
cp.cuda.runtime.setDevice(dev_id)

arr = cp.ones(128, dtype=cp.int64) * (rank + 1)
cp.cuda.get_current_stream().synchronize()

comm.Allreduce(MPI.IN_PLACE, arr)

cp.cuda.Device().synchronize()
assert (arr == (1 + size) * size // 2).all()
```

The error is a `double free or corruption` error. The full error log is displayed below.

```
double free or corruption (out)
[ede90a254643:01264] *** Process received signal ***
[ede90a254643:01264] Signal: Aborted (6)
[ede90a254643:01264] Signal code:  (-6)
double free or corruption (out)
[ede90a254643:01265] *** Process received signal ***
[ede90a254643:01265] Signal: Aborted (6)
[ede90a254643:01265] Signal code:  (-6)
[ede90a254643:01264] [ 0] /usr/lib64/libpthread.so.0(+0x12d10)[0x7f1425b94d10]
[ede90a254643:01265] [ 0] /usr/lib64/libpthread.so.0(+0x12d10)[0x7fa599cacd10]
[ede90a254643:01265] [ 1] [ede90a254643:01264] [ 1] /usr/lib64/libc.so.6(gsignal+0x10f)[0x7fa59918852f]
[ede90a254643:01265] [ 2] /usr/lib64/libc.so.6(gsignal+0x10f)[0x7f142507052f]
[ede90a254643:01264] [ 2] /usr/lib64/libc.so.6(abort+0x127)[0x7fa59915be65]
[ede90a254643:01265] [ 3] /usr/lib64/libc.so.6(abort+0x127)[0x7f1425043e65]
[ede90a254643:01264] [ 3] /usr/lib64/libc.so.6(+0x8f727)[0x7fa5991c9727]
[ede90a254643:01265] [ 4] /usr/lib64/libc.so.6(+0x8f727)[0x7f14250b1727]
[...]

PR #612, opened 2025-02-14
dlpack: Pass capsule around to hold ownership on the managed tensor

dalcinl:
Follow up of #611, cf #609.

cc @seberg @snsunx-nv

### DIFF
--- a/test/arrayimpl.py
+++ b/test/arrayimpl.py
@@ -255,7 +255,8 @@ class BaseDLPackCPU:
 
     def __dlpack__(self, stream=None):
         assert stream is None
-        capsule = dlpack.make_py_capsule(self.array)
+        managed = dlpack.make_dl_managed_tensor(self.array)
+        capsule = dlpack.make_py_capsule(managed, owned=True)
         return capsule
 
     def as_raw(self):
--- a/src/mpi4py/MPI.src/ascaibuf.pxi
+++ b/src/mpi4py/MPI.src/ascaibuf.pxi
@@ -140,15 +140,17 @@ cdef int Py_GetCAIBuffer(object obj, Py_buffer *view, int flags) except -1:
             f"buffer is not contiguous "
             f"(shape:{shape}, strides:{strides}, itemsize:{itemsize})"
         )
+    if readonly and ((flags & PyBUF_WRITABLE) == PyBUF_WRITABLE):
+        raise BufferError(
+            f"__cuda_array_interface__: "
+            f"buffer is not writable"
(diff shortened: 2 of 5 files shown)


test/arrayimpl.py
```diff
@@ -255,7 +255,8 @@ def __dlpack_device__(self):
 
     def __dlpack__(self, stream=None):
         assert stream is None
-        capsule = dlpack.make_py_capsule(self.array)
+        self.managed = dlpack.make_dl_managed_tensor(self.array)
+        capsule = dlpack.make_py_capsule(self.managed)
```
seberg, line 259: Hmmm, ownership of `self.managed` seems dubious here.  Tying both to the `dlmanagedtenesor.ctx` would work, but maybe that is tedious, so how about instead using:
```
Py_INCREF(managed)
PyCapsule_SetContext(managed)
```
and in the destructor decref `PyCapsule_GetContext()` (unconditionally, after dealing with the `managed` destructor).  That way, `managed` lifetime is tied to the capsule (if a bit round-about).

Caching it should also be OK, but if export happens twice the first capsule looks like it would become invalid.  I.e. you have to cache it forever (assuming `array` cannot change that should be OK) and not overwrite it each time.

dalcinl: Please see https://github.com/mpi4py/mpi4py/pull/611#issuecomment-2656191645
Long story short, a capsule destructor implemented as a ctypes callback is unreliable, and it does not work well with exception handling (to the point of segfaulting). The code you are commenting on is only intended for testing within mpi4py, and not a general-purpose thing, I do not expect it to work well outside of mpi4py's testsuite. My primary objective is testing the Cython code and achieving 100% code coverage.

Nonetheless, your comment about catching made me realize I definitely do have lifetime issues in this test code. I'll push a fix shortly.

PS: It is really a pity that capsules do not support weak references, otherwise a weak key dictionary `keepalive[capsule] = managed` would be enough. If you have any leverage in the Python core dev community, you can advocate for such a trivial enhancement.

dalcinl: @seberg Take another look.  As long as I do not trigger exceptions in Cython code, I can setup a capsule destructor with the proper incref/decref to manage lifetime. However, for the tests that involve exceptions, I still have to use borrowed references in the capsule. I guess that's the best I can do with a broken ctypes.

seberg: Ah yeah, sorry... I read it not realizing it was testing, for testing don't worry of course!

PR #684, opened 2025-09-10
fixed erroneous documentation of the TargetSpec

zerothi:
It was typed with TypeSpec, but documentation
showed Datatype.


dalcinl:
Closed via #687. Thanks!

PR #685, opened 2025-09-10
documented None for rc.finalize

zerothi:
It defaults to the `rc.initialize` value.

skip-checks: true


dalcinl:
Closed via #687. Thanks!

PR #690, opened 2025-09-23
test: Handle know failures on Debian

### LINKED ISSUE #688: Disable known test failures on Debian
Let's incorporate skips for known test failures on Debian in mpi4py sources. My preliminary work is this branch: https://github.com/mpi4py/mpi4py/tree/skip-test-debian

I'm looking at the patches here: https://salsa.debian.org/science-team/mpi4py/-/tree/master/debian/patches

I have a minor concerns, though. Looks like the F90 datatypes where not working as expected on PowerPC:
https://salsa.debian.org/science-team/mpi4py/-/blob/master/debian/patches/skip_ppc_failing_tests.patch#L27. 
I tried reproduce the issue runner under a podman container, but I cannot pass `MPI_Init`, I'm getting an UCX error.
Any chance I can get the test output to understand what's going on? I don't see how these tests could fail. 
My only guess is that MPICH's configure is somehow broken in PowerPC, and then is not able to properly detect Fortran types.

Closes #670
Closes #679 
Closes #541  

cc @drew-parsons

dalcinl:
Fixes #688.

cc @drew-parsons

### DIFF
--- a/test/test_datatype.py
+++ b/test/test_datatype.py
@@ -96,6 +96,8 @@ datatypes += datatypes_mpi
 
 combiner_map = {}
 
+machine = platform.machine()
+
 
 class TestDatatypeNull(unittest.TestCase):
     #
@@ -225,7 +227,8 @@ class TestDatatype(unittest.TestCase):
         for dtype in datatypes:
             dtype.Commit()
 
-    @unittest.skipMPI("mpich(<4.2.3)", platform.machine().startswith("ppc"))
+    @unittest.skipMPI("mpich(<4.2.3)", machine.startswith("ppc"))
+    @unittest.skipMPI("openmpi(<6.0.0)", machine in {"ppc", "ppc64"})
     def testCodeCharStr(self):
         f90datatypes = []
         try:
@@ -566,7 +569,8 @@ class BaseTestDatatypeCreateMixin:
         args = (p, r)
         self.check(None, factory, *args)
 
-    @unittest.skipMPI("mpich(<4.2.3)", platform.machine().startswith("ppc"))
+    @unittest.skipMPI("mpich(<4.2.3)", machine.startswith("ppc"))
+    @unittest.skipMPI("openmpi(<6.0.0)", machine in {"ppc", "ppc64"})
     @unittest.skipMPI("openmpi(<3.0.0)")
     @unittest.skipMPI("msmpi")
     def testF90RealDouble(self):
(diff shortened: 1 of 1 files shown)


dalcinl:
I've supplemented the skips with lines like the following:
```
@unittest.skipMPI("openmpi(<6.0.0)", platform.machine() in {"ppc", "ppc64"})
```

drew-parsons:
ppc64 passes with the latest update. Waiting for the last odd architectures to build.

drew-parsons:
mips64el is giving some trouble (with openmpi), sometimes timing out near test_cco_vec `testAlltoallv*`
```
testAlltoallv (test_cco_vec.TestCCOVecWorldDup.testAlltoallv) ... ok
testAlltoallv (test_cco_vec.TestCCOVecWorldDup.testAlltoallv) ... ok
testAlltoallv (test_cco_vec.TestCCOVecWorldDup.testAlltoallv) ... testAlltoallv (test_cco_vec.TestCCOVecWorldDup.testAlltoallv) ... ok
ok
ok
testAlltoallv2 (test_cco_vec.TestCCOVecWorldDup.testAlltoallv2) ... ok
testAlltoallv2 (test_cco_vec.TestCCOVecWorldDup.testAlltoallv2) ... testAlltoallv2 (test_cco_vec.TestCCOVecWorldDup.testAlltoallv2) ... testAlltoallv2 (test_cco_vec.TestCCOVecWorldDup.testAlltoallv2) ... ok
ok
ok
testAlltoallv3 (test_cco_vec.TestCCOVecWorldDup.testAlltoallv3) ... ok
E: Build killed with signal TERM after 300 minutes of inactivity
```
Looking at the historical logs, https://buildd.debian.org/status/logs.php?pkg=mpi4py&arch=mips64el
this appears to be a long-standing issue, not just from the current reorganisation of test skips.

It's been happening all along, it's just that we dealt with it simply by trying the build again. It would eventually pass.

Should some mips64el skips be added to test_cco_vec or should the issue be investigated more closely?
Perhaps it should be reported to openmpi upstream.

dalcinl:
> Looking at the historical logs, https://buildd.debian.org/status/logs.php?pkg=mpi4py&arch=mips64el
this appears to be a long-standing issue, not just from the current reorganisation of test skips.

Also, it seems that the problem is sometimes with `Alltoallv`, and some other times with `Alltoallw`.

> Should some mips64el skips be added to test_cco_vec or should the issue be investigated more closely?

Let's add a skip. 
What's the output of `platform.machine()` here? Is is just `mips64el`? 
Should we use `openmpi(<6.0.0)` as the version bound?

However, I'm not really sure what test or tests to skip. All of the `Alltoallv*` and `Alltoallw*` test methods? 
Maybe this is something you can figure out running in one of these porterbox machines you have access to?

> Perhaps it should be reported to openmpi upstream.

I do not expect them to take any action unless we provide a good, consistent reproducer.

dalcinl:
> There have been package updates in gcc-15

You are getting an assembler error, so the gcc-15 update is probably the culprit.

As temporary workaround is passing `CPPFLAGS=-DPy_LIMITED_API=0x030D0000`. This will build with the CPython 3.13 Stable ABI, thus leading to many different C code paths that may end up skipping the piece that trigger the assembler error.

drew-parsons:
The experts are looking into the sh4 issue, might be a regression in LRA
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1116453

Issue #710, opened 2025-11-01
Intel MPI: Missing key/value pairs from `MPI_Abi_get_fortran_info`


dalcinl:
```python
In [1]: from mpi4py import MPI

In [2]: dict(MPI.Get_abi_fortran_info())
Out[2]: 
{'mpi_logical_size': '4',
 'mpi_integer_size': '4',
 'mpi_real_size': '4',
 'mpi_double_precision_size': '8',
 'mpi_logical1_supported': 'true',
 'mpi_logical2_supported': 'true',
 'mpi_logical4_supported': 'true',
 'mpi_logical8_supported': 'true',
 'mpi_logical16_supported': 'false',
 'mpi_integer1_supported': 'true',
 'mpi_integer2_supported': 'true',
 'mpi_integer4_supported': 'true',
 'mpi_integer8_supported': 'true',
 'mpi_integer16_supported': 'false',
 'mpi_real4_supported': 'true',
 'mpi_real8_supported': 'true',
 'mpi_real16_supported': 'true',
 'mpi_complex8_supported': 'true',
 'mpi_complex16_supported': 'true',
 'mpi_complex32_supported': 'true',
 'mpi_double_complex_supported': 'true'}
```

Two key/value pairs are missing:

* `'mpi_real2_supported': 'false'`
*  `'mpi_complex4_supported': 'false`

cc @alexander-sannikov

dskapoor-intel:
Hello @dalcinl.

I was unable to reproduce the reported behavior. I tried in this environment,

```
# Versions
$ python3.9 --version
Python 3.9.16 :: Intel Corporation 
 
$ python3.9 -m pip show mpi4py | grep -Ei "^(Name|Version):"
Name: mpi4py
Version: 4.1.1

# Query underlying MPI library
$ ldd $(python3.9 -c "import mpi4py.MPI as m; print(m.__file__)") | grep -i libmpi
        libmpi.so.12 => /opt/intel/oneapi/mpi/2021.17/lib/libmpi.so.12 (0x0000126366200000)

# Invocation with mpirun
$ I_MPI_FABRICS=shm  mpirun -n 1 python mpi4py_test_.py
mpi_real2_supported: false
mpi_complex4_supported: false

# Direct invocation
$ I_MPI_FABRICS=shm python mpi4py_test_.py
mpi_real2_supported: false
mpi_complex4_supported: false
```

dalcinl:
[...]

PR #744, opened 2026-04-15
docs: document MPI4PY_LIBMPI environment variable

dskapoor-intel:


docs/source/mpi4py.rst
```diff
@@ -247,6 +247,34 @@ the :data:`mpi4py.rc` and :data:`MPI.pickle` objects at import time of the
...
+  runtime. This is useful on systems with multiple MPI implementations
+  installed, allowing the user to select which MPI library mpi4py should
+  use without modifying system-wide library paths. The value can be an
+  absolute path to the shared library file, or a library filename to be
+  searched in the default library paths.
+
+  Example usage::
+
+    $ export MPI4PY_LIBMPI=/path/to/libmpi.so.12
+    $ export MPI4PY_LIBMPI='$I_MPI_ROOT/lib/libmpi.so.12'
```
david-cortes-intel, line 265: Some comments here:
* Would be helpful to add a link to that variable from the installation instructions for different MPI backends here:
https://mpi4py.readthedocs.io/en/stable/install.html#wheel-packages
* Would be useful to add instructions for Intel MPI specifically for how to get that variable `$I_MPI_ROOT`. It's not obvious to someone unfamiliar with Intel products that one needs to source a script `/opt/intel/oneapi/setvars.sh`.
    * Would also be useful to provide instructions like this for a pip install of `impi_rt` or `openmpi` for example. But not sure how to obtain a path to them programmatically.
* Could also show an analog for windows if it works the same way in that OS.

dskapoor-intel: - I can address the first comment via a link as you suggested.
- I think we should prevent making this look like a getting started with Intel MPI. There are existing references to `$I_MPI_ROOT`. 
- Unfortunately, I haven't tested with Windows and don't have access to a working Windows installation with mpi4py to test this out.

david-cortes-intel (approved):
LGTM, but still ideal if could explicitly tell whether this works on windows or not.

docs/source/mpi4py.rst
david-cortes-intel, line 265: Maybe it shouldn't suggest it down to the .so version number, so that it would be easier for a user to try this.

dskapoor-intel: Actually, this is under example usage. So, it is not meant to be used as-is. It is just an example.
Plus, versioned filenames already appear in the documentation.

dskapoor-intel:
> LGTM, but still ideal if could explicitly tell whether this works on windows or not.

If you have access to a Windows machine to test this out, I can insert a corresponding remark based on your findings.

dalcinl:
@dskapoor-intel Thanks for your PR. I haven't had time to look into it yet, as I said before, I'm quite busy with other stuff and I usually deprioritize documentation-related matters. Your wording is and examples is mostly OK, however I will need to make a few minor tweaks here and there for accuracy. There is also some gotchas related to INSTALL.rst, that file is reused both for the Sphinx documentation and for the PyPI landing page https://pypi.org/project/mpi4py/, therefore I don't think we can use `:envar:` in there. Therefore I need to think how to cross-reference all this stuff. Once I have time, I'll jump in, squash and edit your commits, and hopefully merge.

dalcinl:
@david-cortes-intel @dskapoor-intel I would like to point out that the whole reason users of Intel products may need to use this environment variable is due to Intel Python shiping with outdated copy of `libmpi.so.12`. mpi4py is designed to look in Python directories (site-packages) first, under the principle that custom Python installations and user virtual environments can be made self-contained and independent of what's available at the system level.

IMHO, Intel Python should not ship a copy of this library and rather depend on Intel MPI. I know sometimes things are not that easy and compromises must be made, but maybe there is a better way to do things on your side?

david-cortes-intel:
What do you mean by outdated? The PyPI package `impi_rt` contains the latest version of IMPI:
https://pypi.org/project/impi-rt/#history
https://www.intel.com/content/www/us/en/developer/articles/release-notes/mpi-library-release-notes.html

> mpi4py is designed to look in Python directories (site-packages) first

As far as I can see, the OpenMPI that's uploaded to PyPI is not using `site-packages` either:
https://pypi.org/project/openmpi/#files

.. which makes sense, considering that they aren't really python packages.

> under the principle that custom Python installations and user virtual environments can be made self-contained and independent of what's available at the system level.

Note that there's still nothing preventing a user from installing multiple MPI backends in the same python environment through 'pip'. Or from installing different MPI backends through 'conda' and 'pip', for example.

> but maybe there is a better way to do things on your side?

I do not have any suggestions unfortunately. Perhaps the new wheel-next format could be of help by having a more explicit dependency on MPI, but that's still in an experimental stage.

dalcinl:
> What do you mean by outdated?

The Intel MPI pip packages and conda packages are all good, absolutely no issue with them.

I was talking about the **Intel Python distribution**. I installed it a few weeks ago, while debugging the issue that lead to this PR. IIRC, Intel Python ships with a copy of `libmpi.so.12` from an older version of Intel MPI. This libray copied over from Intel MPI into the Intel Python distribution was somewhat unexpected. 

> > mpi4py is designed to look in Python directories (site-packages) first
> 
> As far as I can see, the OpenMPI that's uploaded to PyPI is not using `site-packages` either

OK, fair enough, I was not accurate in my wording. I should have said `<prefix>/lib`, where `<prefix>` is Python's installation prefix, or the root of a virtual environment.  Exactly this: https://github.com/mpi4py/mpi4py/blob/master/src/mpi4py/_mpiabi.py#L35
I'm the publisher of the MPICH and Open MPI packages. I followed the approach of Intel MPI, that is, installing files under `<prefix>`, and not `<prefix>/lib/pythonX.Y/site-packages`, which is IMO the right thing to do (these are core libraries and not Python modules, as you said).

> Note that there's still nothing preventing a user from installing multiple MPI backends in the same python environment through 'pip'

Well, if they do that, all hell breaks loose, particularly if using `pip`. Files from incompatible packages will simply overwrite on top of each other, and users will endup with a totally broken environment. I don't think we can do better for now. Maybe with wheel-next.

> > but maybe there is a better way to do things on your side?
> 
> I do not have any suggestions unfortunately.

Please note that I was strictly referring to the fact that the **Intel Python distribution** (at least if installed form RPMs) ships with an outdated copy of the Intel MPI shared library. On a second thought, I guess you have nothing to do with the  Intel Python product. Maybe you could at least ping someone from Intel about this matter?

dalcinl:
@david-cortes-intel  Sorry, now I realize you may not have any idea about what I was talking about. The issue with Intel Python shiping a copy of the MPI shared library was discovered here: https://github.com/mpi4py/mpi4py/issues/710#issuecomment-4044574716

david-cortes-intel:
Got it. In my understanding, the IntelPython distribution as an installer is being deprecated in favor of pip/conda packages:
https://www.intel.com/content/www/us/en/developer/articles/technical/get-started-with-intel-distribution-for-python.html

So it shouldn't ship an independent copy of IMPI anymore. Perhaps @jharlow-intel could comment further.

dskapoor-intel:
> @david-cortes-intel @dskapoor-intel I would like to point out that the whole reason users of Intel products may need to use this environment variable is due to Intel Python shiping with outdated copy of `libmpi.so.12`. 

Actually, our motivation behind this PR was different. I suggested this variable to @david-cortes-intel based on his use case that required the use of Intel MPI Library on a system with MPI library from another vendor installed. This variable ensured that mpi4py used Intel MPI. 

Plus, I agree with @david-cortes-intel that based on the IDP documentation Intel MPI Library shouldn't be shipped with IDP. However, we will doublecheck this internally.

PR #757, opened 2026-07-14
build: Bump scikit-build-core requirement and use dynamic-metadata

dalcinl:


dalcinl:
Blocked by codespell-project/codespell#3969

